Issue #31, opened 2015-02-03
Backend crash during pc_patch(pcpoint) or pcpoint::geometry


strk:
~2.3 million points, patched in chunks of 6x6 decimal degrees, triggered a crash in half a minute:
```
strk=# insert into pcpatch24(p) select pc_patch(p) from pcpoint24 group by st_snaptogrid(p::geometry, 6, 6);
The connection to the server was lost. Attempting reset: Failed.
Time: 34290.011 ms
```

strk:
pcpoint::geometry seems immune:
```
select count(*) from ( select st_snaptogrid(p::geometry, 6, 6) from pcpoint24 ) foo;
2736741
```

strk:
NOTE: the pcpatch schema requires GHT compression

Remi-C:
Hey,
could you rule out the postgres memrory issue by executing the exact same
statement but with a classic aggregate instead (need to change your table
definition tough)?

insert into pcpatch24(p) select array_agg(p) from pcpoint24 group by
st_snaptogrid(p::geometry, 6, 6);

It may be faster to round manually:
insert into pcpatch24(p)
select array_agg(p)
from pcpoint24
group by (pc_get(p,'x')/6)::int, (pc_get(p,'y')/6)::int;


I have several billion points in table, with patches up to several
100k, and about a dozen type of points, never crashed (except in my
custom functions).

You can also try to find which patch produce the crash :

first add a unique id per group of points to the table

CREATE TABLE test_crash_pointcloud AS
SELECT * , row_number() over(GROUP BY st_snaptogrid(p::geometry, 6,
6)) as group_id
FROM pcpoint24

Then you can use the pgadmin script mode to loop trough it, printing
the id of group each time




2015-02-03 11:15 GMT+01:00 Sandro Santilli <notifications@github.com>:

Remi-C:
I'm working on it
writting the pgadmin script so you know which patch produces the crash
Cheers,
Rémi-C

2015-02-03 11:51 GMT+01:00 Rémi Cura <remi.cura@gmail.com>:

strk:
Memory is actually low. I just tried to patch a differently compressed (dimensional) set of points, using your conversion strategy, and that one works:
```
[...]

PR #33, opened 2015-02-03
Do not assume pc_patch_ght_from_uncompressed cannot return NULL

strk:
Print an error if it does, and use pcerror from within the function
to reduce the likelihood of returning NULL w/out an actual error.

Fixes the crash reported in #31


strk:
Landed to master as a4df1e54cdd32b40bff235e9a210f1b1fcd0eeea

Issue #47, opened 2015-03-10
PC_Summary(pcpatch)


strk:
Provide a function to inspect a pcpatch reporting various metadata:
 - Number of points
 - Compression type
 - Number of dimensions
 - For each dimension:
    - range (if available in metadata?)
    - compression (for dimensional compression)

List of info to show to be revised

strk:
Continues in #51

PR #51, opened 2015-03-18
First draft of PC_Summary(pcpatch)

strk:
See #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,14 @@ Now that you have created two tables, you'll see entries for them in the `pointc
 >      [-126.93,45.07,7,0],[-126.92,45.08,8,0],[-126.91,45.09,9,0]
 >     ]}
 
+**PC_Summary(p pcpatch)** returns **text** (from 1.1.0)
+
+> Return a JSON formatted summary of the data in that point.
+>
+>     SELECT PC_Summary(pa) FROM patches LIMIT 1;
+>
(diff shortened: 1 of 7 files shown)


strk:
TODO: per-dimension compression (for dimensional) and stats, test and docs

Remi-C:
This is a most useful feature.
Json would be cool because we could use postgres to parse it and get rows if necessary

strk:
json ready for test.

I was tempted to encode dimension as hashes keyed at dimension name, to do things like:

```
select pc_summary(p)::json#>'{dims,X,size}' from mytab
```

but... could a (invalid) schema use the same name for different dimension ?

strk:
oh, the commit also dumps manual string fiddling (death!)

strk:
@pramsey do I need to fully detoast a pcpatch in order to get its dimensional compression info (PCBYTES) ?

pramsey:
Like, what compression is used for each dimension? Yes, that's hiding at the front of each dimension, and they are laid down end-to-end, so you have to read the whole thing.

strk:
uhm, that's unfortunate, for a "summary" function.
Maybe I can pull off only a slice on each dimension end?

strk:
If everything works as expected, the function just told me that the DIMENSIONAL compression of a table resulted in all 8 dimensions using ZLIB compression for one of my cases:
```
strk=# select count(*), (json_array_elements(pc_summary(p)::json->'dims'))->>'compr' compr from (select p from pcpatch8_main limit 1) foo group by compr
;
 count | compr
-------+-------
     8 | zlib
(1 row)
```

And, for all of 101 rows:
```
strk=# select count(*), (json_array_elements(pc_summary(p)::json->'dims'))->>'compr' compr from (select p from pcpatch8_main limit 1) foo group by compr
;
 count | compr
-------+-------
     8 | zlib
(1 row)
```

strk:
On the other hand, a PDAL imported LiDAR, with 13 dimensions:
```
rt=# select count(*), json_array_elements(((pc_summary(p)::json)->'dims'))->>'compr' comp from (select pa p from raw20h64_300 limit 1) foo group by comp;
 count |  comp   
-------+---------
     7 | rle
     4 | zlib
     2 | sigbits
(3 rows)
```

pramsey:
OK, feeling better now

strk:
Squash rebased and merged to master as 3fe9d4599f8fd109d6409b561cea5eac86873d34

PR #55, opened 2015-03-28
Compile the static library with -fPIC

gadomski:
On Ubuntu precise64, gcc/g++ 4.8, cmake 2.8.11.2, I get a link error between the static libpc and the (default shared) libpointcloud.so. This patch fixes the link error.

Fixes #26

PR #56, opened 2015-03-28
Configure cmake files to fix cmake install

gadomski:
CMake install is broken w/o this patch. Depends on #55 (which fixes cmake build).

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -33,6 +33,7 @@ set_target_properties (libpc-static
     OUTPUT_NAME "pc"
     PREFIX "lib"
     CLEAN_DIRECT_OUTPUT 1
+    COMPILE_FLAGS -fPIC
   )
 
 target_link_libraries (libpc-static xml2)
--- a/pgsql/CMakeLists.txt
+++ b/pgsql/CMakeLists.txt
@@ -15,6 +15,16 @@ set ( PC_INSTALL_EXENSIONS
   pointcloud.control
   )
 
+configure_file(
+    pointcloud.sql.in
+    ${CMAKE_CURRENT_LIST_DIR}/pointcloud--${POINTCLOUD_VERSION}.sql
+    )
+
+configure_file(
+    pointcloud.control.in
+    ${CMAKE_CURRENT_LIST_DIR}/pointcloud.control
+    )
+
 set(CMAKE_C_FLAGS "${PGSQL_CPPFLAGS}")
 set(CMAKE_SHARED_LINKER_FLAGS "${PGSQL_LDFLAGS}")
 


strk:
How does install fail for you ? It works for me with cmake 2.8.12.2 as of ac6b91176b41641fdbccdb8b9382023a77cd1ada

Can it be you missed merging #53 in your tree ?

gadomski:
I do have #53 in **install-fix**:

```bash
$ git branch --contains ca0ef74c
  install-fix
* master
```

I double-checked my error, and I do *not* get it on OSX w/ cmake 3.1.3. I only get it on Ubuntu w/ cmake 2.8.11.2.

```
$ sudo make install
[snip]
Install the project...
-- Install configuration: "RelWithDebInfo"
-- Installing: /usr/lib/postgresql/9.1/lib/pointcloud.so
CMake Error at pgsql/cmake_install.cmake:68 (FILE):
  file INSTALL cannot find
  "/home/vagrant/pointcloud/pgsql/pointcloud--1.1.0.sql".
Call Stack (most recent call first):
  cmake_install.cmake:38 (INCLUDE)
[snip]
$ ls /home/vagrant/pointcloud/pgsql/
CMakeFiles  CMakeLists.txt  expected  Makefile  META.json  pc_access.c  pc_inout.c  pc_pgsql.c  pc_pgsql.h  pointcloud.control.in  pointcloud.sql.in  sql
```

On OSX `pointcloud--1.1.0.sql` and `pointcloud.control` are happily generated alongside their `*.in` parents. How are those files generated? I can't find a corresponding `configure_file`. Is cmake doing some magical file configuration when `pointcloud--1.1.0.sql` is `install`d?

As an aside, as I was digging through the history to find #53, I noticed that you have two mostly symmetric branches going in `pgpointcloud/pointcloud`, **master** and **1.0.x**. May I ask why you are creating separate commits in each branch for what appear to be the same patches? Seems like duplication?

strk:
> On OSX `pointcloud--1.1.0.sql` and `pointcloud.control` are happily generated alongside their `*.in` parents. How are those files generated? I can't find a corresponding `configure_file`. Is cmake doing some magical file configuration when `pointcloud--1.1.0.sql` is `install`d?

Uhm, I was wrong. It looks like I did have .sql files in the source tree
due to an autoconf-based build, that's why it worked.

Pushed to master (07af590f6b6e4d19e2999509aee868633bae72dc)
and 1.0.x (8b48017b1aed0cd3b365629db0fb2db04757c9ec) branches.

> As an aside, as I was digging through the history to find #53, I noticed that you have two mostly symmetric branches going in `pgpointcloud/pointcloud`, **master** and **1.0.x**. May I ask why you are creating separate commits in each branch for what appear to be the same patches? Seems like duplication?

1.0.x is a bugfix only branch.
master targets 1.1.0

I know it's anoying to have different commit hashes for the same patch
but the I think the only way to avoid that would be merging a branch
based at a common ancestor, which is not necessarely a straightforward
thing to setup.

--strk;

gadomski:
Fair enough, though I don't think it's too hard to do something like this:

bugfix

This lets you share bugfix commits between branches and (IMO) makes it clearer what's a bugfix and what's a new feature. But no worries, just was a bit confusing to figure out at first, but I think I got it now.

PR #84, opened 2015-05-15
Implement PC_Patch{Min,Max,Avg}(PcPatch) returns PcPoint

### LINKED ISSUE #77: PC_Patch Min/Max/Avg return PcPoint
Companion to PcPatchMin/Max/Avg taking a dimension name, these functions would not take a dimension name but rather return all of them as a PcPoint. It would be the fastest way to retrieve all stats from a single patch (which would otherwise require decompressing patches once per dimension).

strk:
Closes #77

NOTE: changes the C function signature so you need to properly
      upgrade scripts for things to work again.


strk:
Merged as 38944e72dfeaff3595d78033eb2d0cb25aa5e5ad

PR #85, opened 2015-05-26
add support for binary protocol for pcpatch

cyrilleberger:


strk:
Great! Would you be able to also provide a testcase for this ?

cyrilleberger:
I will have a look. My test cases for it are deeply entangled in our application, but I should be able to write something standalone. Do you have existing testcase for the SQL function that can I have a look at?

strk:
The SQL: https://github.com/pgpointcloud/pointcloud/tree/master/pgsql/sql
The expected results: https://github.com/pgpointcloud/pointcloud/tree/master/pgsql/expected

Tests are run by "make installcheck", you can see them run by travis here:
https://travis-ci.org/pgpointcloud/pointcloud/builds

Testing binary input/output is tricky. In PostGIS we do it via COPY, see
https://github.com/postgis/postgis/blob/svn-trunk/regress/binary.sql

Remi-C:
Hm some newbee question : what are the binary protocol used for?
Cheers

pramsey:
When defined with the 'send' and 'recv' slots, they determine what bytes will be sent when a CURSOR is declared as BINARY. Oftentimes sending data over a binary cursor is more efficient than sending over a text one, though it can sometimes be hard to write a good client since you have to deal with the binary representations of *all* types, not just the big ones you "care" about.

strk:
it'll be fun to try at finding a ways to allow upgrades w/out dump-reload...

pramsey:
Can't you? As long as on-disk format doesn't change, there's no reason you cannot use pg_upgrade...

strk:
upgrades of pointcloud, not postgresql. there's no way to ALTER TYPE for adding the recv and send attributes.

cyrilleberger:
Actually there is a way, you can do:
UPDATE pg_type SET typsend = 'pcpatch_send' WHERE typname='pcpatch';
UPDATE pg_type SET typrecv = 'pcpatch_recv' WHERE typname='pcpatch';

cyrilleberger:
So the challenge is that pg_regress use the text protocol. In fact, looking into postgres and postgis code source, it seems that they don't test the binary protocol at all. What I would need to do is write a small C program that run the test.

strk:
PostGIS regression testing does not use pg_regress but its own invocation
of `psql`. The COPY .. BINARY construct I pointed you to earlier does
the trick, no need to write C.

Thanks for the UPDATE calls, hopefully that's not too much of a moving
target across PostgreSQL versions. The current way to update is about just
copying the enabler script to the upgrade script (skipping some components
that are known to not be upgradeable) so putting those UPDATEs in the 
pointcloud.sql.in file should do it (to be tested). Note there's a
pc_upgrade() script that you can use to import changes even within the
same in-development version.

strk:
This PR is still in need of properly encoding the UPGRADE path (making sure `ALTER EXTENSION update` properly inject the new send/recv elements of the type)

elemoine:
This is old and inactive. Closing…

PR #88, opened 2015-07-30
Fix PG 9.5 compatibility.

vpicavet:
Applied this PostGIS patch to pgpointcloud : https://trac.osgeo.org/postgis/changeset/13797/branches/2.0
I do not think we support PG 8.4 for pgpointcloud, hence no test.
Tested against PG9.5 pgdg repositories


strk:
Successfully tested against PG 9.3.4.
Guess we need to test with 9.1 too

strk:
PG 9.1 actually fails installcheck, but I'm not sure it's related: #89

Issue #30, opened 2015-02-02
Confusing pcpatch::geometry cast


strk:
pcpoint::geometry returns a POINT geometry, pcpatch::geometry returns a POLYGON geometry, shouldn't it be a MULTIPOINT instead ? I guess it's returning the envelope instead, but sounds confusing.

strk:
also, is there a way to get a multipoint geometry out of a pcpatch ?

pramsey:
Not currently (well, explode, cast those points to geometry, and collect).

Remi-C:
Hm,
the interest of pg_pointcloud would be very limited if the cast to geom was
to multipoint.

For instance a multimillion point patch would be very very big in geometry.

The idea is precisely to "generalize" a multipoint with an area.

One of my concern with this is that I find this area a little bit too much
geenralized.
FOr instance an orietned bounding box would be much better for me
(or even a surface looking closer to original points, I wrote a function
for this, whihc is quite expensive)

Cheers,
Rémi-C

2015-02-02 17:22 GMT+01:00 Paul Ramsey <notifications@github.com>:

strk:
To add to the confusion, an index built on geometry(pcpatch) is not used when the query filter uses pc_envelope(pcpatch) instead. But as noted in this issue the geometry cast is really nothing different from a call to pc_envelope(pcpatch), so why the duplication ? In other words, why should the "generalization" take the form of a cast ?

strk:
answering to myself, I just realized that pc_envelope returns a bytea rather than a geometry, and the planner uses a cast from bytea to geometry when it comes to compare in a spatial filter, which does not match the field used for creating the index.

mbredif:
I also find the semantics of casting with pc_envelope not so intuitive.
I would rather have a cast to multipoint and indices explicitly built using pc_envelope.
Would that be viable or would it incur some performance penalty?
I understand that pc_envelope returns a hand built bytea to prevent a c-dependency on postgis. Would not it be more intuitive to have pgpointcloud expose the current function as something like pc_envelope_aswkb, and have the pointcloud-postgis SQL extension expose pc_envelope as a POLYGON geometry?

PR #96, opened 2016-02-17
memcpy size seems erroneous for u_int32_t : 8 -> 4

mbredif:
(not tested)

### DIFF
--- a/pgsql/pc_pgsql.c
+++ b/pgsql/pc_pgsql.c
@@ -908,7 +908,7 @@ pc_patch_wkb_set_double(uint8_t *wkb, double d)
 static uint8_t *
 pc_patch_wkb_set_int32(uint8_t *wkb, uint32_t i)
 {
-	memcpy(wkb, &i, 8);
+	memcpy(wkb, &i, 4);
 	wkb += 4;
 	return wkb;
 }


strk:
Sounds good, but please add a test.

mbredif:
Hi,
`pc_patch_wkb_set_int32` is not defined in the header file, so I guess I cannot test it directly. It appears that it is only used in `pc_patch_to_geometry_wkb_envelope` where the "bug" does not matter much as `int32`s are written in sequence and followed by a `pc_patch_wkb_set_double` (l.976), so the writing is well defined.  I do not know the exact spec of `memcpy`, but attempting to read 8 bytes from an 4-byte int does not sound good to me anyhow.

strk:
reading 8 bytes from an 4-byte int is undefined behavior. the app would mostly read random 4 bytes from somewhere, which may or may not be assigned to the process (if it's not it'll be a segmentation fault).

Valgrind should be able to report such invalid read, so even if a test doesn't properly _fail_, it could be still useful to check it via Valgrind, and require it to run valgrind-clean (something that could be added to the travis check, btw).

mbredif:
While trying to get `valgrind` notice that issue, I ended up moving the envelope wkb construction from `pgsql` to `lib` as a  ̀pc_bounds_to_wkb` function, leaving only a serialization relevant code in `pgsql`.

The issue is that  ̀make installcheck` is more of a black box than `make check` so that embedding it and PostgreSQL within valgrind is beyond my reach. I was able to move the code and provide a cunit test, but I could not get valgrind to notice the wrong size parameter.

@strk, do you mind if I package that within the same PR as the disambiguation discussed in #30 :
- PC_Envelope(pcpatch) : wkb -> PC_Envelope_AsBinary(pcpatch) : wkb
- geometry(pcpatch) : geometry -> PC_Envelope(pcpatch) : geometry

mbredif:
Superseded by #104.

PR #97, opened 2016-02-23
fix crash when a child node is empty in pc schema

ldgeo:
A segmentation fault occurs and crash postgresql when a child node is empty in a PC schema definition.

This PR tries to fix that. 

Tested only against PostgreSQL 9.5.

### DIFF
--- a/lib/cunit/cu_pc_schema.c
+++ b/lib/cunit/cu_pc_schema.c
@@ -54,6 +54,20 @@ test_schema_from_xml()
 }
 
 static void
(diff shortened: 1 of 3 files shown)


strk:
May I suggest to move the "emptyfieldschema" and support variables in the single method using it, rather than keeping it global ?

strk:
Could you please rebase to current master, so to also get automated valgrind testing in place ?
Feel free to also squash commits as you go.

pblottiere:
ok, it's done!

strk:
Thanks!

PR #101, opened 2016-03-02
valgrind-clean

mbredif:
steps towards addressing #100 : 
- [x] fix memory leak in test_uncompressed_filter (0826a986a0fcbc812f37c6bdf45fe6aea88c4d1f)
- [x] issuing missing xmlCleanupParser calls (40b2e49d894581dcc4c191a917f97a6c30c031f3)
- [x] Conditional jump or move depends on uninitialised value(s) / Use of uninitialised value of size 4 / Syscall param write(buf) points to uninitialised byte(s) in  `pc_double_to_ptr (pc_val.c:182)` (897d7c086d6dcd931b8d75395a3a3dcec983982b)
- [x] reachable lost block : call_init (dl-init.c:30). This still reachable lost block is not of the concern of pgpointcloud code and may be disregarded -> nothing to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: c
 
 before_install:
   - sudo apt-get update
-  - sudo apt-get install -q postgresql-server-dev-9.3 libcunit1-dev
+  - sudo apt-get install -q postgresql-server-dev-9.3 libcunit1-dev valgrind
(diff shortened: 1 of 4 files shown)


strk:
Thanks, excellent work

PR #108, opened 2016-06-16
Consistent error handling in pc_schema_from_xml

achidlow:
Hello!

This is my first ever pull-request on GitHub, so if I've done something incorrectly, please let me know.

This small patch is mostly to ensure that pc_schema_from_xml returns PC_FAILURE when pc_schema_check_xy fails. This was causing issues whilst I was trying to wrap this lib idiomatically with Cython.

Also, whilst I was doing the above, rather than just copy and paste the cleanup code yet again, I decided to use the standard C error handling pattern and put all the cleanup in one place for the sake of legibility and consistency.


achidlow:
Oh, didn't realise there was unit tests. Looks like CMake doesn't generate a `make test` target? I'll fix the failing unit test

achidlow:
Closing the request until I get some other issues sorted out.

achidlow:
I'v fixed the issue I found when using the library externally, which turned out was two uninitialized variables in pc_schema_from_xml.

I've added a unit test for this case, and also expanded the from_xml and clone unit tests.

Remi-C:
Hey, I also tried (and failed miserably) to wrap the lib in cython
would be most interested to here about your experience with it.
Cheers

2016-06-16 8:34 GMT+02:00 Adam Chidlow <notifications@github.com>:

achidlow:
@Remi-C I'm hoping my employer will let me open source the bindings*. It's a work in progress but going well so far. The only challenge really has been to make errors more Pythonic. I've installed a custom error handler that throws an Exception, but since that is in a callback Cython requires a signal on the method return value for when to check for Exceptions. It's better performance/cleaner code if methods that can call `pcerror()` and which return `PC_SUCCESS` or `PC_FAILURE`, do so consistently.

(*If my employer doesn't let me open source the bindings I'd be happy to discuss the experience / offer advice over email.)

Remi-C:
Thanks for the anwser, this would be perfect !


Much point cloud processing/handling is coming to python nowadays,
having an open source official python wrapper of pgpointcloud would greatly
increase its value for your employer,
or so I hope ;-)

I would be dearly curious to benchmark the cython version vs the (very
partial) pure python version regarding parsing and cast to numpy array.

Please do not hesitate to keep us posted about your progress !

Cheers,
Rémi-C

2016-06-17 7:37 GMT+02:00 Adam Chidlow <notifications@github.com>:

achidlow:
@Remi-C I've got parsing `PCPATCH` to/from a numpy array working and as fast as I can reasonably make it. Do you know where I could find the pure python version to compare against?

Remi-C:
Hey Adam,
I wrote a pure python class to parse a (uncompressed) pcpatch into a numpy
array (array with specialised type or plain double float array)
https://github.com/Remi-C/PPPP_utilities/blob/master/pointcloud/pg_pointcloud_classes.py
You can see an example of how it works in the function "test_from_outside_db"
(L473),
The main function is "WKB_patch_to_numpy_double" (L247).

you can use it from inside (plpython) or outside (psycopg),
because the class needs to retrieve the pcschema (which is then cached in a
python global dictionary).

I tried to make it as fast as I could without resorting to trick
detrimental to simplicity,

I'm very curious to see how much faster your C wrapping will be (I'd say at
least 10 times?)

Cheers,
Rémi-C


2016-06-23 4:08 GMT+02:00 Adam Chidlow <notifications@github.com>:

hobu:
> I wrote a pure python class to parse a (uncompressed) pcpatch into a numpy
array (array with specialised type or plain double float array)

Is there value in making lazperf accessible from Python? I've kind of wanted this, but I didn't know if anyone else would actually use it.

mbredif:
Now that #146 is merged, this one may be closed.

Issue #111, opened 2016-06-21
memory leak in pc_patch_compress


achidlow:
Found a leak by running valgrind on my python unit tests.

It's because `pc_patch_free` is what frees the PCSTATS but `pc_patch_dimensional_free` is called instead.

```
==25371== 114 (72 direct, 42 indirect) bytes in 1 blocks are definitely lost in loss record 618 of 1,663
==25371==    at 0x4C29C0F: malloc (vg_replace_malloc.c:299)
==25371==    by 0x154CF34A: default_allocator (pc_mem.c:40)
==25371==    by 0x154CF5EE: pcalloc (pc_mem.c:136)
==25371==    by 0x154D4DBD: pc_stats_clone (pc_stats.c:133)
==25371==    by 0x154D08DF: pc_patch_dimensional_from_uncompressed (pc_patch_dimensional.c:87)
==25371==    by 0x154CFBBF: pc_patch_compress (pc_patch.c:142)
==25371==    by 0x154B0C0E: __pyx_pf_4pgpc_7PcPatch_10compressed (pgpc.c:7595)
==25371==    by 0x154B0C0E: __pyx_pw_4pgpc_7PcPatch_11compressed (pgpc.c:7574)
==25371==    by 0x53EE15: PyEval_EvalFrameEx (in /usr/bin/python3.5)
==25371==    by 0x53F1F3: PyEval_EvalFrameEx (in /usr/bin/python3.5)
==25371==    by 0x5449FA: PyEval_EvalCodeEx (in /usr/bin/python3.5)
==25371==    by 0x4E7F66: ??? (in /usr/bin/python3.5)
==25371==    by 0x5C3676: PyObject_Call (in /usr/bin/python3.5)
==25371==    by 0x53CC34: PyEval_EvalFrameEx (in /usr/bin/python3.5)
==25371==    by 0x5449FA: PyEval_EvalCodeEx (in /usr/bin/python3.5)
==25371==    by 0x4E7E52: ??? (in /usr/bin/python3.5)
==25371==    by 0x5C3676: PyObject_Call (in /usr/bin/python3.5)
==25371==    by 0x4DEB5D: ??? (in /usr/bin/python3.5)
==25371==    by 0x5C3676: PyObject_Call (in /usr/bin/python3.5)
==25371==    by 0x56C265: ??? (in /usr/bin/python3.5)
==25371==    by 0x5C3676: PyObject_Call (in /usr/bin/python3.5)
[...]

PR #112, opened 2016-06-21
Fix pc_patch_compress leak

achidlow:
This fixes a memory leak in pc_patch_compress. See issue #111.

### DIFF
--- a/lib/cunit/cu_pc_patch.c
+++ b/lib/cunit/cu_pc_patch.c
@@ -364,6 +364,57 @@ test_patch_dimensional_compression()
 }
 
 static void
+test_patch_compression_stats_leak()
+{
+    PCPOINT *pt;
+    int i;
+    int npts = 400;
+    PCPOINTLIST *pl1, *pl2;
+    PCPATCH *pch1, *pch2;
+    PCDIMSTATS *pds = NULL;
+
+    pl1 = pc_pointlist_make(npts);
+
+    for ( i = 0; i < npts; i++ )
+    {
+        pt = pc_point_make(simpleschema);
+        pc_point_set_double_by_name(pt, "x", i*2.0);
+        pc_point_set_double_by_name(pt, "y", i*1.9);
+        pc_point_set_double_by_name(pt, "Z", i*0.34);
+        pc_point_set_double_by_name(pt, "intensity", 10);
+        pc_pointlist_add_point(pl1, pt);
+    }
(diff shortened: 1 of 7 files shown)


mbredif:
I came across this issue as well :
- `pc_patch_dimensional_free(pcdu);` is not equivalent (as one could expect) with `pc_patch_free((PCPATCH*)pcdu);` because it does not free the stats (https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_patch.c#L85-L89).

They could be made equivalent, saving a upcast-switch-downcast round-trip by putting the stats-freeing code (possibly wrapped in a function such as `pc_patch_common_free(PCPATCH*)` ) in all the `pc_patch_*_free` specialized functions.

achidlow:
@mbredif that sounds like a better idea, that would fix any other similar leaks in one go. I'll get on it.

achidlow:
Okay I've made the change @mbredif suggested and made a separate `pc_patch_common_free` function. This actually revealed an error in `pc_patch_dimensional_compress` where the stats member was set by `memcpy` from another `PCPATCH_DIMENSIONAL` resulting in a double-free. So I went through every case of `pcalloc(sizeof(PCPATCH*` to make sure the `stats` member was being set correctly, either to `NULL` or through `pc_stats_clone`

strk:
Is there an existing test in the testsuite that would make the leak evident (and confirm the fix) ?

mbredif:
I have just [stumbled](https://travis-ci.org/mbredif/pointcloud/builds/197587351) yet again on this issue while addressing #135 in [mbredif/pointcloud/pointlist_data](/mbredif/pointcloud/tree/pointlist_data).

The non-equivalence of `pc_patch_*_free(pa)` with `pc_patch_free((PCPATCH*)pa);` is so easy oversee...

I included a [quick fix](https://github.com/mbredif/pointcloud/blob/pointlist_data/lib/pc_patch_uncompressed.c#L230) but this PR would be more relevant.

elemoine:
> The non-equivalence of pc_patch_*_free(pa) with pc_patch_free((PCPATCH*)pa); is so easy oversee...

That's right, but couldn't the rule be "never call `pc_patch_[something]_free`, and always call `pc_patch_free`"?

elemoine:
Replaced by https://github.com/pgpointcloud/pointcloud/pull/206.

PR #117, opened 2016-09-11
Update README.md

BenoitDalFerro:
Hi,

First wonderful job, pointcloud is a very nice piece of code, but I guess you already know that.

I did this change because as a Windows user that came across your early papers I could not find how to install it from github until I discovered it was already shipped with postGIS.

I therefore added a warning to your readme so that people don't lose an evening scratching their heads like I did.

All the best,

Benoit

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@ A PostgreSQL extension for storing point cloud (LIDAR) data.
 
 ## Build/Install ##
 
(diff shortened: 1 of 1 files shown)


strk:
To my knowledge it is NOT "shipped with PostGIS". But I don't hold knowledge of Windows packages.
Maybe @robe2 might know better

elemoine:
Closing this. @BenoitDalFerro feel free to reopen if we're missing the point.

Issue #118, opened 2016-10-01
Master not compiling against PostgreSQL 9.6


pjduplooy:
Hi guys.  

I am trying to compile Master against PostgreSQL 9.6, and am experiencing some compile problems.

PostgreSQL was installed from the pgdg apt.
Postgis, laz-perf and libght compiled from source.
All other deps installed from repositories. (Ubuntu 16.04.1 and osgeolive ppa for gdal and others)

The errors I get are:

[ 90%] Building C object pgsql/CMakeFiles/pointcloud.dir/pc_access.c.o
In file included from /usr/include/postgresql/9.6/server/access/heapam.h:22:0,
                 from /usr/include/postgresql/9.6/server/nodes/execnodes.h:18,
                 from /usr/include/postgresql/9.6/server/executor/execdesc.h:18,
                 from /usr/include/postgresql/9.6/server/executor/executor.h:17,
                 from /usr/include/postgresql/9.6/server/funcapi.h:21,
                 from /home/pieter/Temp/pointcloud/pgsql/pc_access.c:14:
/usr/include/postgresql/9.6/server/storage/lock.h:18:2: error: #error "lock.h may not be included from frontend code"
 #error "lock.h may not be included from frontend code"
  ^
In file included from /usr/include/postgresql/9.6/server/storage/lock.h:23:0,
                 from /usr/include/postgresql/9.6/server/access/heapam.h:22,
                 from /usr/include/postgresql/9.6/server/nodes/execnodes.h:18,
                 from /usr/include/postgresql/9.6/server/executor/execdesc.h:18,
                 from /usr/include/postgresql/9.6/server/executor/executor.h:17,
                 from /usr/include/postgresql/9.6/server/funcapi.h:21,
                 from /home/pieter/Temp/pointcloud/pgsql/pc_access.c:14:
/usr/include/postgresql/9.6/server/storage/lwlock.h:18:2: error: #error "lwlock.h may not be included from frontend code"
 #error "lwlock.h may not be included from frontend code"
  ^
In file included from /usr/include/postgresql/9.6/server/storage/lwlock.h:22:0,
                 from /usr/include/postgresql/9.6/server/storage/lock.h:23,
                 from /usr/include/postgresql/9.6/server/access/heapam.h:22,
                 from /usr/include/postgresql/9.6/server/nodes/execnodes.h:18,
                 from /usr/include/postgresql/9.6/server/executor/execdesc.h:18,
[...]

PR #119, opened 2016-10-19
Fix compile against PostgreSQL 9.6 on Debian by using correct flags.

achidlow:
Fixes issue #118.

pgpointcloud is C, it should be using CLFAGS not CPPFLAGS.


pramsey:
Thanks, makes sense.

Issue #122, opened 2017-01-23
Build issues when using a specific install prefix


elemoine:
I am trying to build pointcloud with both GHT and LAZPERF, using cmake. The thing is that I installed GHT and LAZPERF in a specific directory (`~/local`), and it seems that the pointcloud build system does not support that.

This is the error I get:

```bash
$ mkdir build
$ cd build
$ cmake -DCMAKE_INSTALL_PREFIX=~/local ../
…
$ make
[  3%] Building CXX object lib/CMakeFiles/liblazperf-static.dir/lazperf_adapter.cpp.o
In file included from /home/elemoine/src/pointcloud/lib/pc_api_internal.h:18:0,
                 from /home/elemoine/src/pointcloud/lib/lazperf_adapter.hpp:12,
                 from /home/elemoine/src/pointcloud/lib/lazperf_adapter.cpp:10:
/home/elemoine/src/pointcloud/lib/pc_api.h:25:17: fatal error: ght.h: No such file or directory
 #include "ght.h"
                 ^
compilation terminated.
lib/CMakeFiles/liblazperf-static.dir/build.make:62: recipe for target 'lib/CMakeFiles/liblazperf-static.dir/lazperf_adapter.cpp.o' failed
make[2]: *** [lib/CMakeFiles/liblazperf-static.dir/lazperf_adapter.cpp.o] Error 1
CMakeFiles/Makefile2:89: recipe for target 'lib/CMakeFiles/liblazperf-static.dir/all' failed
make[1]: *** [lib/CMakeFiles/liblazperf-static.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

After fixing the `#include "ght.h"` issue I encounter new problems at `cu_tester` and `pointcloud` link time.

I'll try to come up with a PR that fixes these issues.

PR #123, opened 2017-01-23
Fix cmake build issues

elemoine:
This pull request fixes https://github.com/pgpointcloud/pointcloud/issues/122. Currently the cmake-based build system doesn't handle the case where GHT and LAZPERF are installed in a specific directory (outside `/usr`).

Also, I am wondering if it makes sense to support both cmake and autotools. Wouldn't it make sense to ditch one of the two systems? If so, which one should stay?


strk:
Travis is using autotools to test, so I would not ditch that one.
That said, I trust your judgment on the need for this PR, which looks fine although I hadn't tested it.

elemoine:
> Travis is using autotools to test, so I would not ditch that one.

I am currently working on making the Travis script work with cmake instead of autotools. I am happy to submit a PR if you are (like me) interested in ditching autotools.

strk:
I'm happy if Travis build swith both (in a matrix).
Ditching autotools is not particularely interesting for me at the moment.

elemoine:
Ok up to you. For your information, this is a PR in our fork that removes autotools: https://github.com/LI3DS/pointcloud/pull/27. We just think that it does not make too much sense to maintain both systems, but you may have arguments…

strk:
No specific arguments other than last I tried it failed :)
On other projects I've found that cmake builds lacked libtool scripts
to test things before installing them (which needed to be replaced
by custom things). In other cases I found myself with my OS packaged
cmake version being too old to be able to build. Finally (least important)
I know autotools better than cmake.

The above are my points, but that said I think the least used system
will automatically vanish.

PR #124, opened 2017-01-23
Fix unit tests with cunit 2.1-3

elemoine:
The `cu_tester` binary currently segfaults on Debian 9 (Sid). This is because Debian 9 has CUnit 2.1-3, in which the `CU_SuiteInfo` structure has changed.

This PR changes the unit test code in a way that it works with both CUnit 2.1-2 and 2.1-3.


strk:
Thanks for this, looks great from here :)

PR #126, opened 2017-01-25
Use __attribute__ ((format (printf, x, y)))

elemoine:
This commit adds `__attribute__ ((format (printf, x, y)))` where it makes sense.

This removes the following compile warning:

```
[ 96%] Building C object pgsql/CMakeFiles/pointcloud.dir/pc_pgsql.c.o
/home/elemoine/src/pointcloud/pgsql/pc_pgsql.c: In function ‘pgsql_msg_handler’:
/home/elemoine/src/pointcloud/pgsql/pc_pgsql.c:66:2: warning: function might be possible candidate for ‘gnu_printf’ format attribute [-Wsuggest-attribute=format]
  vsnprintf(msg, MSG_MAXLEN, fmt, ap);
  ^~~~~~~~~
```

The compilation is warning-free after this for me!

I hesitated to create a macro for this, so that we don't use `__attribute__` when the compiler is not GNUC. But I am not sure pgpointcloud actually supports other compilers.

### DIFF
--- a/lib/pc_api.h
+++ b/lib/pc_api.h
@@ -25,6 +25,10 @@
 #include "ght.h"
 #endif
 
+#ifndef __GNUC__
+#define __attribute__ (x)
+#endif
+
 /**********************************************************************
 * DATA STRUCTURES
 */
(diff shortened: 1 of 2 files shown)


strk:
In PostGIS we put this somewhere included by all __attribute__ users:
```
#ifndef __GNUC__
# define __attribute__ (x)
#endif
```

elemoine:
Yep I can do that.

elemoine:
> In PostGIS we put this somewhere included by all attribute users

Done.

strk:
And... merged! Thanks

PR #128, opened 2017-01-27
Remove unused functions and vars

elemoine:
When using autotools (as opposed to cmake) the `-Wunused-variable` flag is used, revealing that a number of functions and variables are not used. This PR removes the unused functions and variables.

There's actually one left in `pc_patch.c`:

```
gcc -Wall -O2 -g -fPIC -I/usr/include/libxml2  -I/home/elemoine/local/include -I/home/elemoine/local/include/  -c -o pc_patch.o pc_patch.c                                                                         
pc_patch.c: In function ‘pc_patch_compress’:                                                                                                                                                                       
pc_patch.c:248:21: warning: unused variable ‘pal’ [-Wunused-variable]                                                                                                                                              
    PCPATCH_LAZPERF *pal = pc_patch_lazperf_from_uncompressed((PCPATCH_UNCOMPRESSED*)patch);                                                                                                                       
                     ^~~
```

But this one is more serious, I think it requires an actual fix for an actual bug.

### DIFF
--- a/lib/cunit/cu_pc_bytes.c
+++ b/lib/cunit/cu_pc_bytes.c
@@ -58,13 +58,8 @@ static PCBYTES initbytes(uint8_t *bytes, size_t size, uint32_t interp)
 static void
 test_run_length_encoding()
 {
-	char *bytes, *bytes_rle, *bytes_de_rle;
+	char *bytes;
 	int nr;
-	uint32_t bytes_nelems;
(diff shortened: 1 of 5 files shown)


elemoine:
> But this one is more serious, I think it requires an actual fix for an actual bug.

Addressed by https://github.com/pgpointcloud/pointcloud/pull/130.

Issue #57, opened 2015-03-28
Provide a "check" target for cmake based build


strk:
Neither "make test" nor "make check" worked on a cmake build tree.
It's needed.

pblottiere:
Considering that the cmake support has been dropped (#202), this issue can be closed.

PR #129, opened 2017-01-27
Add check and installcheck targets to CMake-based Makefile

### LINKED ISSUE #18: How to test under CMake
Can't for the life of me figure out how to make check under cmake.  I asked and you said you aren't sure it's doable, so am ticketing.

elemoine:
Fixes #18 #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ Make a build directory, and run ``cmake`` from there. Use ``ccmake`` to browse a
 - ``make``
 - ``sudo make install``
 
+Run unit tests:
+
+- ``make check``
+
 #### Autotools ####
 
 After generating the configure script with ``autogen``,  ``./configure --help`` to get a complete listing of configuration options.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,6 +46,13 @@ After generating the configure script with ``autogen``,  ``./configure --help``
 - ``make``
 - ``sudo make install``
 
+Run unit tests:
+
+- ``make check``
+
+Note that if you configured PointCloud using a non-standard LibGHT location, you may need to add its location to the ``LD_LIBRARY_PATH`` environment variable. For example:
```
strk: This should not be needed IFF test binaries are created via libtool (which produces scripted wrappers to do just this).
Do you confirm this is not the case for our tests ? And, does this only apply to the autotools method ?

elemoine: Yes, this is based on actual tests. And yes, it only occurs when using Autotools. CMake uses rpath or something and does not suffer the same problem.

PR #131, opened 2017-01-30
Fix indent in cu_pc_sort.c

elemoine:
cu_pc_sort.c currently uses a mix of spaces and tabs. This commit changes all the tabs to spaces.


strk:
It would be good to drop an .editorconfig file in the root of the project.
See http://editorconfig.org/

PR #134, opened 2017-02-01
Build and test with libght in Travis

elemoine:
This PR suggests changing the Travis script to build and test pointcloud with both lazperf and libght.

Unfortunately the Valgrind check reports many memory issues when pointcloud is compiled with HAVE_LIBGHT defined. And the installcheck target fails as well. So I suggest that we run valgrind and installcheck without libght for now.

### DIFF
--- a/.install-libght.sh
+++ b/.install-libght.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+set -ex
+git clone https://github.com/pramsey/libght.git
+cd libght; cmake .; make; sudo make install; sudo ldconfig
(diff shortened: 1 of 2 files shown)


Remi-C:
I'm not sure of the status of pgpointcloud with light right now.
Last time I heard it was more or less bugged.
Cheers

2017-02-01 16:19 GMT+01:00 Éric Lemoine <notifications@github.com>:

elemoine:
> I'm not sure of the status of pgpointcloud with light right now. Last time I heard it was more or less bugged.

I can at least report that Valgrind doesn't like it too much, probably for good reasons.

This PR just makes the Travis script build pointcloud with libght, and also run the unit tests with libght (which currently pass). So this is a first step towards improving the situation.

elemoine:
Closing this for now. We can re-open if people care about libght.

PR #113, opened 2016-08-03
fix laz support (memory issue and lack in lazperf type)

pblottiere:
This PR fixes three issues in laz support:
- the lack of support for PC_INT64, PC_UINT64, PC_DOUBLE, and PC_FLOAT for laz compression
- a memory error due to a too soon free in the lifecycle of a PCPOINTLIST coming from a lazperf patch
- a pgsql expected output for laz tests

PR #135, opened 2017-02-01
Change pc_pointlist_from_uncompressed implementation

elemoine:
Currently `pc_pointlist_from_uncompressed` creates readonly points, that is points whose data is stored in the uncompressed patch's data block.

First, this is inconsisent with the behavior of [`pc_pointlist_from_dimensional`](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_pointlist.c#L61-L92), which creates non-readonly points (using the `pc_point_make` function).

This further means that the high-level `pc_pointlist_from_patch` is not consistent, because it returns readonly points for some patch types and non-readonly points for others.

Also, [`pc_pointlist_from_lazperf`](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_patch_lazperf.c#L71-L79) and [`pc_pointlist_from_ght`](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_patch_ght.c#L574-L580) are currently implemented in terms of `pc_pointlist_from_uncompressed`. The former frees the intermediary uncompressed patch before returning the pointlist, which is not correct because this makes the points point to `free`'d memory. The latter does not free the uncompressed patch, causing a memory leak.

This patch makes `pc_pointlist_from_uncompressed` create non-readonly points instead of readonly points. In this way, `pc_pointlist_from_patch` is consistent, always returning non-readonly points. And it makes it possible to implement `pc_pointlist_from_lazperf` and `pc_pointlist_from_ght` in terms of `pc_pointlist_from_uncompressed`.

When (if) this is merged, [`pc_pointlist_from_ght`](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_patch_ght.c#L574-L580)  will still need to be changed, to free the intermediary uncompressed buffer. This will be done with a separate PR.

This patch is an alternative to https://github.com/pgpointcloud/pointcloud/pull/113, which just fixes `pc_pointlist_from_lazperf` by using `pc_point_make` instead of relying on `pc_pointlist_from_uncompressed`.

Please review.

### DIFF
--- a/lib/pc_api.h
+++ b/lib/pc_api.h
@@ -314,6 +314,9 @@ PCPOINT* pc_point_make(const PCSCHEMA *s);
 /** Create a new readonly PCPOINT on top of a data buffer */
 PCPOINT* pc_point_from_data(const PCSCHEMA *s, const uint8_t *data);
 
+/** Copy the data to the point data, based on the point size */
+void pc_point_copy_data(PCPOINT *pt, const uint8_t *data);
+
 /** Create a new read/write PCPOINT from a double array */
 PCPOINT* pc_point_from_double_array(const PCSCHEMA *s, double *array, uint32_t nelems);
 
--- a/lib/pc_point.c
+++ b/lib/pc_point.c
@@ -64,6 +64,11 @@ pc_point_from_data(const PCSCHEMA *s, const uint8_t *data)
 	return pt;
 }
 
+void
+pc_point_copy_data(PCPOINT *pt, const uint8_t *data)
+{
+    memcpy(pt->data, data, pt->schema->size);
+}
 
 void
 pc_point_free(PCPOINT *pt)
--- a/lib/pc_pointlist.c
+++ b/lib/pc_pointlist.c
@@ -102,7 +102,9 @@ pc_pointlist_from_uncompressed(const PCPATCH_UNCOMPRESSED *patch)
 	pl = pc_pointlist_make(npoints);
 	for ( i = 0; i < npoints; i++ )
 	{
-		pc_pointlist_add_point(pl, pc_point_from_data(patch->schema, patch->data + i*pt_size));
+		PCPOINT *pt = pc_point_make(patch->schema);
+		pc_point_copy_data(pt, patch->data + i * pt_size);
+		pc_pointlist_add_point(pl, pt);
 	}
 	return pl;
 }


mbredif:
Good find. Your proposal actually fixes the issue but removes the optimization that prevents heap allocation when it is possible to return a readonly pointlist.
My proposal is to transfer ownership from the to-be-deleted data from the uncompressed patch to the pointlist. The idea is implemented [here](/mbredif/pointcloud/tree/pointlist_data) but might need some polish.

mbredif:
Here is the more polished version : 
-  mbredif/pointcloud@1774b9c replaces the unused `readonly` attribute with an opaque `mem` memory pointer owned by the pointlist (ie: which should be freed when the pointlist is freed). the implementation is streamlined using a new [`void *pc_patch_uncompressed_free_mem (PCPATCH_UNCOMPRESSED *patch)`](https://github.com/mbredif/pointcloud/blob/1774b9c86346a906248e4fece5f4e91a7b72bff8/lib/pc_patch_uncompressed.c#L225-L233) function.
- mbredif/pointcloud@396442d2 uses that [poinlist-wide shared `mem`](https://github.com/mbredif/pointcloud/blob/396442d25d2b79b49cd2df4595cf0b7d08366aed/lib/pc_pointlist.c#L77) pointer rather than individual data pointers in each point to reduce the number of mallocs from 1 per point to 1 per patch.

elemoine:
> Your proposal actually fixes the issue but removes the optimization that prevents heap allocation when it is possible to return a readonly pointlist.

That's a valid point.



> Here is the more polished version :

I provided comments in your commits. I think there's a memory leak in your implementation.

mbredif:
I replied in your comments there (mbredif/pointcloud@1774b9c, mbredif/pointcloud@396442d )
mbredif/pointcloud@4e57f27 fixes the memory leak.
What do you think between this PR and [mbredif/pointcloud/pointlist_data](/pgpointcloud/pointcloud/compare/master...mbredif:pointlist_data) ?

mbredif:
By the way, this PR invalidates the LI3DS/pointcloud#29, LI3DS/pointcloud#41 and similar upcoming PRs.

mbredif:
Based on your input @elemoine, I prepared the alternative PR #136.

elemoine:
Closing this in favor of #136.

PR #136, opened 2017-02-03
pointlist memory fix, pc_pointlist_from_dimensional memory optimization

mbredif:
@elemoine , @strk, this is an alternative PR to #135, with cleaner semantics (I think).

- The readonly attribute of pointlist was never used. I propose to change it for an opaque memory pointer `mem`, that will point to the data owned by the pointlist, if any so that it can be freed upon destruction.

- The new function `pc_patch_uncompressed_readonly` makes the patch readonly and return the disowned `data`, if any, to enable transfer ownership (eg: affectation to the `pointlist->mem` pointer

- As a bonus, pc_pointlist_from_dimensional can now be implemented with a single pcalloc/patch rather than point.

### DIFF
--- a/lib/pc_api.h
+++ b/lib/pc_api.h
@@ -128,7 +128,7 @@ typedef struct
 
 typedef struct
 {
-	int8_t readonly;
+	void *mem; /* An opaque memory buffer to be freed on destruction if not NULL */
 	uint32_t npoints;
 	uint32_t maxpoints;
 	PCPOINT **points;
--- a/lib/pc_patch_uncompressed.c
+++ b/lib/pc_patch_uncompressed.c
@@ -222,6 +222,15 @@ pc_patch_uncompressed_free(PCPATCH_UNCOMPRESSED *patch)
 	pcfree(patch);
 }
 
+// Make the patch readonly. Return the memory segment
+// owned by the patch, if any, to enable transfer of ownership
+uint8_t *
+pc_patch_uncompressed_readonly(PCPATCH_UNCOMPRESSED *patch)
+{
+	uint8_t *data = patch->readonly ? NULL : patch->data;
+	patch->readonly = PC_TRUE;
+	return data;
+}
 
 
 PCPATCH_UNCOMPRESSED *
--- a/lib/pc_api_internal.h
+++ b/lib/pc_api_internal.h
@@ -180,6 +180,7 @@ PCPATCH_UNCOMPRESSED* pc_patch_uncompressed_make(const PCSCHEMA *s, uint32_t max
 int pc_patch_uncompressed_compute_extent(PCPATCH_UNCOMPRESSED *patch);
 int pc_patch_uncompressed_compute_stats(PCPATCH_UNCOMPRESSED *patch);
 void pc_patch_uncompressed_free(PCPATCH_UNCOMPRESSED *patch);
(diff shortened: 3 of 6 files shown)


lib/pc_api.h
```diff
@@ -126,7 +126,7 @@ typedef struct
 
 typedef struct
 {
-	int8_t readonly;
+	void *mem; /* An opaque memory buffer to be freed on destruction if not NULL */
```
elemoine: I would remove the comment here, as it pertains to the internals and not the API.

lib/pc_patch_uncompressed.c
```diff
@@ -222,6 +222,15 @@ pc_patch_uncompressed_free(PCPATCH_UNCOMPRESSED *patch)
 	pcfree(patch);
 }
 
+// Makes the patch readonly. Returns the memory segment
```
elemoine: (nit) `s/Makes/Make`

lib/pc_patch_uncompressed.c
```diff
@@ -222,6 +222,15 @@ pc_patch_uncompressed_free(PCPATCH_UNCOMPRESSED *patch)
...
+// owned by the patch, if any, to enable transfer of ownership
+uint8_t *
+pc_patch_uncompressed_readonly(PCPATCH_UNCOMPRESSED *patch)
+{
+	uint8_t *data = patch->readonly ? NULL : patch->data;
+	patch->readonly = PC_TRUE;
+	return data;
+}
```
elemoine: I would personally rewrite this as this:

```c
uint8_t *
pc_patch_uncompressed_readonly(PCPATCH_UNCOMPRESSED *patch)
{
    unit8_t *data = NULL;
    if (patch->readonly == PC_FALSE)
    {
        data = patch->data;
        patch->readonly = PC_TRUE;
    }
    return data;
}
```

Up to you, just a suggestion.

mbredif:
Rebased after #138, Please review.

This fixes a [memory leak](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_patch_ght.c#L574-L580) and a [return value referencing free'd memory](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_patch_lazperf.c#L71-L79)

elemoine:
> Rebased after #138, Please review.

FWIW, these changes make sense to me, and I think this PR should be merged.

I just had a few minor comments that you may want to consider @mbredif.

strk:
Please also rebase it, while you're at it

lib/pc_api.h
mbredif: I see, not sure where to relocate it in pc_pointlist.c though.

lib/pc_patch_uncompressed.c
mbredif: My version is more concise and makes it clear that the output patch is readonly. But that is a matter of taste. I can go for your version if required.

mbredif:
Rebased

lib/pc_patch_uncompressed.c
elemoine: This was just a suggestion. It's a matter of taste. Keep your version if you like it best.

strk:
Is this good to merge, then ?
I'm relying on peer-reviews here (not adding my review..)

mbredif:
Good for me!

elemoine:
@strk this looks good to me, a good patch that fixes actual bugs.

strk:
Thanks for testing, merged.

PR #130, opened 2017-01-27
Make pc_patch_compress return the compressed patch

elemoine:
When compressing a patch from `GHT` to `LAZPERF` the pc_patch_compress function currently returns the uncompressed patch. This patch fixes the problem.

This patch removes this warning:

```
pc_patch.c: In function ‘pc_patch_compress’:                                                                                                                                                                       
pc_patch.c:248:21: warning: unused variable ‘pal’ [-Wunused-variable]                                                                                                                                              
    PCPATCH_LAZPERF *pal = pc_patch_lazperf_from_uncompressed((PCPATCH_UNCOMPRESSED*)patch);                                                                                                                       
                     ^~~
```

PR #140, opened 2017-02-13
Remove compile warning

elemoine:
When using `./configure CFLAGS="-Wall -O2 -g"` we get an "unused-function" warning at compile time. This was raised by @strk in #130. This commit fixes the issue.


strk:
thanks !

PR #141, opened 2017-02-17
Remove tabs within if statements

elemoine:
This is a fix style only patch.


pblottiere:
Thank you for cleaning!

PR #142, opened 2017-02-18
pc_point_to_geometry_wkb : get_m fix

mbredif:
A typo was introduced while rebasing this PR: `z` values were assigned to both `z` and `m` output points


strk:
Pushed, thanks.
This means that code path is lacking a testcase, right ?
Up to cover it ?

mbredif:
ok : #143 is ready for review

PR #144, opened 2017-03-08
Add z and m to PCBounds

elemoine:
I am creating this PR for discussion, on behalf of @mbredif and myself.

Currently, the [bounds](https://github.com/pgpointcloud/pointcloud/blob/60e10e31ef7114f7819456f4f9ed065e10278013/lib/pc_api.h#L147-L153) that we store in patches are 2D. This is a limitation.

For example, `PC_Intersects(p1, p2)`, which checks whether the bounds of `p1` and `p2` intersect, works in [2D only](https://github.com/pgpointcloud/pointcloud/blob/60e10e31ef7114f7819456f4f9ed065e10278013/lib/pc_util.c#L238-L249). Also, `geometry(p)` currently [returns](https://github.com/pgpointcloud/pointcloud/blob/60e10e31ef7114f7819456f4f9ed065e10278013/pgsql_postgis/pointcloud_postgis--1.0.sql#L16-L26) a 2D polygon. So with an index on `geometry(p)` only 2D spatial relationship tests are possible.

To address these limitations this PR suggests adding `zmin`, `zmax`, `mmin` and `mmax` to `PCBOUNDS`.

One issue is the change to the patch format, making "old" patch data incompatible with newer versions of pointcloud. I think this is not acceptable. One way to address that problem might be to use some bits of the [current patch header](https://github.com/pgpointcloud/pointcloud/blob/60e10e31ef7114f7819456f4f9ed065e10278013/lib/pc_api.h#L172-L178) to store some version number. For example, we could steal some bits from `type`, which only requires [4 bits](https://github.com/pgpointcloud/pointcloud/blob/60e10e31ef7114f7819456f4f9ed065e10278013/lib/pc_api.h#L36-L45) at the moment.

What do you think? Do you agree with extending `PCBOUNDS`? Does the idea of stealing bits from `type` to encode some version in the data sound valid? Stupid? Do you see other way to address the "patch data compatibility" issue?

### DIFF
--- a/lib/pc_api.h
+++ b/lib/pc_api.h
@@ -150,6 +150,10 @@ typedef struct
 	double xmax;
 	double ymin;
 	double ymax;
+	double zmin;
+	double zmax;
+	double mmin;
+	double mmax;
 } PCBOUNDS;
 
 /* Used for generic patch statistics */
--- a/lib/pc_api_internal.h
+++ b/lib/pc_api_internal.h
@@ -269,11 +269,13 @@ void pc_bytes_to_ptr(uint8_t *buf, PCBYTES pcb, int n);
 */
 
 /** Initialize with very large mins and very small maxes */
-void pc_bounds_init(PCBOUNDS *b);
+void pc_bounds_init(PCBOUNDS *b, const PCSCHEMA *schema);
 /** Copy a bounds */
 PCSTATS* pc_stats_clone(const PCSTATS *stats);
 /** Expand extents of b1 to encompass b2 */
 void pc_bounds_merge(PCBOUNDS *b1, const PCBOUNDS *b2);
+/** Expand extents of b to encompass p */
+void pc_bounds_expand(PCBOUNDS *b, const PCPOINT *p);
 
 /****************************************************************************
 * BITMAPS
--- a/lib/pc_filter.c
+++ b/lib/pc_filter.c
@@ -165,7 +165,7 @@ pc_patch_dimensional_filter(const PCPATCH_DIMENSIONAL *pdl, const PCBITMAP *map)
 		stats.max = pc_value_scale_offset(stats.max, dim);
 		stats.sum = pc_value_scale_offset(stats.sum, dim);
 
-		/* Save the X/Y stats for use in bounds later */
+		/* Save the XYZM stats for use in bounds later */
 		if ( i == pdl->schema->x_position )
 		{
 			fpdl->bounds.xmin = stats.min;
@@ -175,6 +175,16 @@ pc_patch_dimensional_filter(const PCPATCH_DIMENSIONAL *pdl, const PCBITMAP *map)
 		{
 			fpdl->bounds.ymin = stats.min;
 			fpdl->bounds.ymax = stats.max;
+		}
+		else if ( i == pdl->schema->z_position )
+		{
+			fpdl->bounds.zmin = stats.min;
+			fpdl->bounds.zmax = stats.max;
+		}
+		else if ( i == pdl->schema->m_position )
+		{
+			fpdl->bounds.mmin = stats.min;
+			fpdl->bounds.mmax = stats.max;
 		}
 
 		pc_point_set_double_by_index(&(fpdl->stats->min), i, stats.min);
--- a/lib/pc_patch_dimensional.c
+++ b/lib/pc_patch_dimensional.c
@@ -21,7 +21,7 @@ typedef struct
 	int8_t readonly;
 	const PCSCHEMA *schema;
 	uint32_t npoints;
-	double xmin, xmax, ymin, ymax;
+	double xmin, xmax, ymin, ymax, zmin, zmax, mmin, mmax;
 	PCSTATS *stats;
 	PCBYTES *bytes;
 } PCPATCH_DIMENSIONAL;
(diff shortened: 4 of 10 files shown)


pgsql/pc_pgsql.c
```diff
@@ -530,9 +530,10 @@ pc_patch_dimensional_serialize(const PCPATCH *patch_in)
 {
 	//  uint32_t size;
 	//  uint32_t pcid;
-	//  uint32_t compression;
+	//  uint16_t version;
```
mbredif: Version and compression should be swapped here and in all similar comments, to reflect the ordering in the declaration.

pgsql/pc_pgsql.c
```diff
@@ -799,16 +806,43 @@ pc_patch_serialize_to_uncompressed(const PCPATCH *patch_in)
 	return serpatch;
 }
 
+static void
+pc_patch_set_common(PCPATCH *patch, const SERIALIZED_PATCH *serpatch, const PCSCHEMA *schema)
```
mbredif: I propose to rename this as `pc_patch_common_deserialize`.
It should return somehow the data pointer to the payload, since we cannot rely on serpatch->data for patches in version 0.

pgsql/pc_pgsql.c
```diff
@@ -799,16 +806,43 @@ pc_patch_serialize_to_uncompressed(const PCPATCH *patch_in)
...
+		// (xmin, xmax, ymin, ymax, zmin, zmax, mmin, mmax)
+		patch->type = serpatch->compression;
+		patch->bounds = serpatch->bounds;
+	}
+	else
+	{
+		// old-style header, with 2D bounds
+		// (xmin, xmax, ymin, ymax)
+		patch->type = (uint16_t) compression_and_version;
+		memcpy(&patch->bounds, &serpatch->bounds, 4 * sizeof(double));
```
mbredif: z and m bounds should be initialized with [-inf,inf]

pgsql/pc_pgsql.c
```diff
@@ -799,16 +806,43 @@ pc_patch_serialize_to_uncompressed(const PCPATCH *patch_in)
...
+	patch->npoints = serpatch->npoints;
+
+	compression_and_version = *((uint32_t *) (&serpatch->compression));
+	if ( compression_and_version > UINT16_MAX)
+	{
+		// new-style header, with 4D bounds
+		// (xmin, xmax, ymin, ymax, zmin, zmax, mmin, mmax)
+		patch->type = serpatch->compression;
+		patch->bounds = serpatch->bounds;
+	}
```
mbredif: return `serpatch->data` ?

pgsql/pc_pgsql.c
```diff
@@ -799,16 +806,43 @@ pc_patch_serialize_to_uncompressed(const PCPATCH *patch_in)
...
+	else
+	{
+		// old-style header, with 2D bounds
+		// (xmin, xmax, ymin, ymax)
+		patch->type = (uint16_t) compression_and_version;
+		memcpy(&patch->bounds, &serpatch->bounds, 4 * sizeof(double));
+	}
```
mbredif: return serpatch->data-4*sizeof(double) ?

pgsql/pc_pgsql.c
```diff
@@ -550,6 +551,7 @@ pc_patch_dimensional_serialize(const PCPATCH *patch_in)
 	serpch->pcid = patch->schema->pcid;
```
mbredif: Some code could be factored from all patch types into a new `pc_patch_common_serialize` function. (This function could eventually allow serializing version 0 serpatches. is there any point to it ?)

pgsql/pc_pgsql.c
elemoine: Agree.

elemoine:
> My main concern is that serpatch header changed so that its new data field will not point to the beginning of the payload for patches in version 0.

That part is not yet clear to me. I guess I need to study the code.

elemoine:
> My main concern is that serpatch header changed so that its new data field will not point to the beginning of the payload for patches in version 0.

Obviously. I understand the issue. I am now wondering whether using two different structures would be better: `SERIALIZED_PATCH_V0` and `SERIALIZED_PATCH_V1`. This would also mean introducing `PCBOUNDS4` instead of extending the current `PCBOUNDS`. I'll give it a try.

mbredif:
`SERIALIZED_PATCH` are in multiple places...
Considering that the `bounds` is in the `data` (similar to the `stats`) would be easier to implement. Then it would require a mechanism to get the bounds, stats and payload offsets within the data pointer (similar to the functionnality currently offered by [pc_stats_size](https://github.com/pgpointcloud/pointcloud/blob/master/lib/pc_stats.c#L190))

pgsql/pc_pgsql.c
elemoine: Done.

pgsql/pc_pgsql.c
elemoine: I am trying to minimize changes for now.

pgsql/pc_pgsql.c
elemoine: Done.

pgsql/pc_pgsql.c
elemoine: Done.

pgsql/pc_pgsql.c
elemoine: Done.

pgsql/pc_pgsql.c
elemoine: Done. Using `DBL_MAX`.

elemoine:
@mbredif I finally continued with the initial approach. And I followed your suggestions to have `pc_patch_common_serialize` return a pointer to the serpatch data.

mbredif:
Good ! There are however be other places where `compression` was used that need further modifications (the list might not be complete...) : 
- `compression` variables are still declared as `uint32_t` in various places ([like here](https://github.com/pgpointcloud/pointcloud/blob/04993240d921cf5a049bcf8f6b8624183f4669dd/lib/pc_patch.c#L132-L133) or [in PCSCHEMA](https://github.com/pgpointcloud/pointcloud/pull/144/files#diff-d7c7bdde4dd00cc24459067092b2844fR95))
- [wkb_get_compression](https://github.com/pgpointcloud/pointcloud/blob/692a5681012b635222e15858766cc2ba8484cb42/lib/pc_util.c#L176) needs a fix

elemoine:
> There are however be other places where compression was used that need further modifications (the list might not be complete...) :

I know that the patch is incomplete. This is WIP. At this point, I am just trying out things and discussing options with you :) (Thanks a lot for that!)

> compression variables are still declared as uint32_t in various places (like here or in PCSCHEMA)

Yep.

> wkb_get_compression needs a fix

I am not sure about this. Why do you think `wkb_get_compression` requires a fix exactly? The compression is [represented](https://github.com/LI3DS/pointcloud/blob/a6c4b6b9e16e723077a03ce864ae769ac32e0d85/lib/pc_patch_uncompressed.c#L69-L93) as an `uint32_t` in the WKB string, and I see no reason to change that. Changing the compression type in the WKB would mean breaking existing data/applications again.


Another thing that certainly requires changes: [`pc_patch_to_geometry_wkb_envelope`
](https://github.com/LI3DS/pointcloud/blob/a6c4b6b9e16e723077a03ce864ae769ac32e0d85/pgsql/pc_pgsql.c#L1064-L1129)

That function is used by `PC_Envelope`, itself used by the pointcloud_postgis [`geometry`](https://github.com/LI3DS/pointcloud/blob/a6c4b6b9e16e723077a03ce864ae769ac32e0d85/pgsql_postgis/pointcloud_postgis--1.0.sql#L19-L26) to create a PostGIS polygon from a patch. It needs to return a 3D envelope to be able to do 3D relationship tests with an index on `geometry(p)`.

mbredif:
Ok I see your point of not breaking everything. A possibility would be to steal bytes from patch wkb as well, keeping a wkb version of 0, so that compression is uint16 everywhere.

For the envelope, I propose to align with postgis as much as possible. [ST_Envelope](http://postgis.net/docs/ST_Envelope.html) is a 2D polygon (despite the mention in the documentation)

> (PostGIS will add a ZMIN/ZMAX coordinate as well)


```
# SELECT Box3D(geom), Box2D(geom), ST_AsText(ST_Envelope(geom)) As envelope 
 FROM (SELECT 'MULTIPOINT((0 0 0), (3 4 10))'::geometry As geom) As foo;
        box3d        |    box2d     |            envelope            
---------------------+--------------+--------------------------------
 BOX3D(0 0 0,3 4 10) | BOX(0 0,3 4) | POLYGON((0 0,0 4,3 4,3 0,0 0))
```

[Here](https://github.com/LI3DS/pointcloud/pull/59/commits/f044097eae5703f655b757bc32e4e486e8e29a15) is an attempt to provide a box3d.

@strk, @pramsey : what would be the preferred way to expose the 3DZ, 3DM and 4D bounding boxes of patches so that it integrates well with postgis and postgresql in general (multidimensional indexes) ? From a quick look at postgis source code, box3D seems to get obsolete and GBOX is not exposed in sql.

elemoine:
> Ok I see your point of not breaking everything. A possibility would be to steal bytes from patch wkb as well, keeping a wkb version of 0, so that compression is uint16 everywhere.

Yes, I thought about that. But this also means changes that we don't really need. I don't know right now.



> For the envelope, I propose to align with postgis as much as possible. ST_Envelope is a 2D polygon

Good point on `ST_Envelope` always returning a 2D geometry. I hadn't realized that.

elemoine:
@mbredif see https://github.com/pgpointcloud/pointcloud/pull/144/commits/ed807c686f29ae4a462b21d57ba1089c154904e2 (in my `pcbounds-xyzm` branch) for a slightly different way to identify new vs old-style patches. In this approach the `compression` field remains an `uint32_t`, but an offset is applied at serialization time. This is actually not very different from the previous patch, but by still using a `uint32_t` for `compression` we keep the consistency with the type used in the WKB.

strk:
@mbredif I'd use a bounding diagonal: http://postgis.net/docs/ST_BoundingDiagonal.html

mbredif:
> @mbredif I'd use a bounding diagonal: http://postgis.net/docs/ST_BoundingDiagonal.html
Ok, sounds good, thanks for the pointer.

elemoine:
@mbredif if you have a chance to look at my patch again…

mbredif:
I think your proposal does the job, the last version doing the strict minimum amount of change. My concern is that whatever new format ends up being decided , it will have to be supported, so that the new format should be good and generic enough so that we do not have to break it tomorrow. I have no experience in format maintenance though...

elemoine:
> I think your proposal does the job, the last version doing the strict minimum amount of change. My concern is that whatever new format ends up being decided , it will have to be supported, so that the new format should be good and generic enough so that we do not have to break it tomorrow. I have no experience in format maintenance though...

Yep, I have the same concern as you. No rush. Let's keep thinking (and experimenting) about this… Thanks for your feedback.

pramsey:
Main comment is this is an invasive change for something that in 90% of cases won't help. At least in large scale LIDAR world, no? I found all the selectivity was in x/y, and most queries were taking in the majority of the range of z. Having full dimensional indexes is something that sounds like a good idea but doesn't help that much. If you have a use case where you **need** it, then by all means ignore me, but if you just think it **seems sensible** then take a pause.

elemoine:
@pramsey yes, I agree, and that's what I meant by "let's keep thinking about the whole thing…" It is not my plan to merge this.

elemoine:
We have other ideas to explore, so I am closing this for now. Thank you all for your comments.

PR #145, opened 2017-03-09
Make it possible to have multiple SQL files

elemoine:
This PR introduces minor changes to the pgsql `CMakeLists.txt` and `Makefile` files. These changes have no impact on the pointcloud build, but they ease the maintainance of our pointcloud version by making it possible to have multiple SQL files in the pointcloud extension.


elemoine:
I am abandoning this PR. After discussion with @pblottiere and @mbredif we don't want to introduce a change to the pointcloud build system just for the needs of a separate project. We need to find a better way to extend pointcloud in that separate project.

PR #146, opened 2017-03-10
Fix bug in pc_patch_filter for lazperf

elemoine:
In `pc_patch_filter`, when `nset` is `0`, the empty patch returned was created with an invalid maxpoints value in the case of lazperf.

### DIFF
--- a/lib/cunit/cu_pc_patch_lazperf.c
+++ b/lib/cunit/cu_pc_patch_lazperf.c
@@ -246,6 +246,41 @@ test_wkb_lazperf()
 	pcfree(wkb1);
 	pcfree(wkb2);
 }
+
+static void
+test_patch_filter_lazperf_zero_point()
+{
+	PCPOINT *pt;
+	int i;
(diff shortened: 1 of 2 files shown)


lib/cunit/cu_pc_patch_lazperf.c
```diff
@@ -246,6 +246,41 @@ test_wkb_lazperf()
 	pcfree(wkb1);
 	pcfree(wkb2);
 }
+
+static void
+test_patch_filter_lazperf()
```
elemoine: I'd suggest using a more explicit name for the test function. Like `test_patch_filter_lazperf_zero_points`.

pblottiere:
I renamed the test to `test_patch_filter_lazperf_zero_points` as you suggested.

pblottiere:
I think this PR is ready to be merged.

elemoine:
Thanks for renaming the test function. Merging.

PR #148, opened 2017-03-15
Change non-breaking space to regular space

elemoine:
This tiny patch changes a non-breaking space character to a regular space character in `configure.ac`.


pblottiere:
It's ok for me. Thank you @elemoine

PR #149, opened 2017-03-17
PC_EnvelopeAsBinary -> bytea, PC_Envelope -> geometry

mbredif:
Following #30 and #86, I propose this simple PR concerning the SQL api.

It basically renames `PC_Envelope` as `PC_EnvelopeAsBinary`, which properly states what it does, leaving room to rename `geometry(pcpatch)` as `PC_Envelope(pcpatch)`.

The cast PCPATCH::geometry is removed since it is ambiguous with the 3D or 4D geometry exports (3D envelope, some 4D geometry to give access to the 4D bbox...) that have been made possible by the recent support for ZM dimensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,8 @@ Now that you have created two tables, you'll see entries for them in the `pointc
 
 ## Functions ##
 
+### PcPoint Functions
+
 **PC_MakePoint(pcid integer, vals float8[])** returns **pcpoint**
 
 > Given a valid `pcid` schema number and an array of doubles that matches the schema, construct a new `pcpoint`.
@@ -238,14 +240,6 @@ Now that you have created two tables, you'll see entries for them in the `pointc
 > 
 >     1     
 
-**PC_AsBinary(p pcpoint)** returns **bytea**
-
-> Return the OGC "well-known binary" format for the point.
->
->     SELECT PC_AsBinary('010100000064CEFFFF94110000703000000400'::pcpoint);
->
->     \x01010000800000000000c05fc000000000008046400000000000005f40
-
 **PC_Get(pt pcpoint, dimname text)** returns **numeric**
 
 > Return the numeric value of the named dimension. The dimension name
@@ -262,6 +256,8 @@ Now that you have created two tables, you'll see entries for them in the `pointc
 >     SELECT PC_Get('010100000064CEFFFF94110000703000000400'::pcpoint);
 >
 >     {-127,45,124,4}
+
+### PcPatch Functions
 
 **PC_Patch(pts pcpoint[])** returns **pcpatch**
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -276,12 +276,12 @@ Now that you have created two tables, you'll see entries for them in the `pointc
 > 
 >     1     
  
-**PC_Envelope(p pcpatch)** returns **bytea**
+**PC_Envelope_AsBinary(p pcpatch)** returns **bytea**
```
elemoine: I'd remove this function completely. If I want to get an OGC WKB representation of the envelope polygon I can just do:

```SQL
select ST_AsBinary(PC_Envelope(patch));
```

We don't need PointCloud to provide a specific function for that.

README.md
```diff
@@ -519,6 +519,11 @@ The `pointcloud_postgis` extension adds functions that allow you to use PostgreS
 > 
 >     POINT Z (-127 45 124)
 
+**PC_Envelope(pcpatch)** returns **geometry**<br/>
```
elemoine: Let's make the doc more verbose, as done for other functions. We can reuse the "Useful for performing intersection tests with geometries" statement I guess.

README.md
```diff
@@ -519,6 +519,11 @@ The `pointcloud_postgis` extension adds functions that allow you to use PostgreS
...
+**pcpatch::geometry** returns **geometry**
```
elemoine: The `geometry` function and cast were not documented before. Given that we do not plan to encourage their use I would not add them to the doc.

README.md
mbredif: How would you implement that without linking to postgis library?

README.md
mbredif: Agreed. To be removed.

README.md
mbredif: Ok. We should emphasize as well that it is only 2d.

README.md
elemoine: I am just suggesting to have `PC_Envelope` in the pointcloud-postgis extension, exactly the way you did in this PR. And if the user needs a WKB representation of the patch envelope then he will have to use `PC_Envelope` and PostGIS' `ST_AsBinary` function. So, no, no linking to the PostGIS lib.

README.md
elemoine: Yes. We should explicitly say that the return value is a PostGIS 2D Polygon representing a box that bounds the patch along the X and Y axis. Feel free to change the wording though :)

README.md
elemoine: Sorry, I realize this is not very clear. I would just not expose and advertise `PC_Envelope_AsBinary` in the doc but I understand that we need this function for the implementation of `PC_Envelope`. Sorry for not being clear.

mbredif:
I have just updated the proposal. I removed the pcpatch::geometry cast from the docs and the sql, with a note in the Readme explaining its absence and regrouped the PC_AsBinary and PC_Enveloppe_AsBinary in a OGC WKB section.

README.md
elemoine: I take that you disagree with no exposing `PC_Envelope_AsBinary` to the user?

mbredif: It feels weird to remove PC_Envelope_AsBinary and keep PC_AsBinary, which is more or less the same  case. Should we remove it as well?

elemoine: Ok, I get your point. I don't think it's very useful but we can keep it. I'd just suggest to rename it to `PC_EnvelopeAsBinary` for consistency reasons. Neither PointCloud nor PostGIS uses a mix of CamelCase and Underscore notations.

mbredif:
Great. `PC_Envelope_AsBinary` function is now renamed to `PC_EnvelopeAsBinary`.

elemoine:
@pblottiere do you want to look at that? This looks good to me.

elemoine:
> I removed the pcpatch::geometry cast from the docs and the sql

I realize that this breaks the compatibility. I would not remove the `geometry` cast and function. I would not add them to the doc but I would not remove them from the SQL file.

mbredif:
would you then remove the comment https://github.com/pgpointcloud/pointcloud/pull/149/files#diff-04c6e90faac2675aa89e2176d2eec7d8R536?

elemoine:
> would you then remove the comment https://github.com/pgpointcloud/pointcloud/pull/149/files#diff-04c6e90faac2675aa89e2176d2eec7d8R536?

Yes.

elemoine:
@mbredif I think you can squash your two commits into one.

mbredif:
squashed!

elemoine:
Sorry. This needs a rebase.

mbredif:
Rebased !

PR #150, opened 2017-03-18
Fixes and improvements to the README file

elemoine:
This PR suggests a number of fixes and improvements to the documentation.


pblottiere:
Thank you for your work @elemoine!

I think a good move would be to update the PDAL pipeline too. In fact, it's JSON now, not XML anymore.

elemoine:
> I think a good move would be to update the PDAL pipeline too. In fact, it's JSON now, not XML anymore.

Ok, I will add a commit for that.

elemoine:
New commit added. And I added another one for the Build Status icon.

The generated page can be seen here: https://github.com/elemoine/pointcloud/blob/readme/README.md.

pblottiere:
Thank you for updating the README @elemoine! It looks good to me!

PR #152, opened 2017-03-23
Add PC_Range(p PCPATCH, first int4, count int4)

elemoine:
This commit adds a `PC_Range` function. This function returns a patch of "count" points. These points are selected from the start-th point in the input patch.

Patch originally written by @pblottiere.

### DIFF
--- a/lib/pc_patch.c
+++ b/lib/pc_patch.c
@@ -306,7 +306,7 @@ PCPATCH *
 pc_patch_from_wkb(const PCSCHEMA *s, uint8_t *wkb, size_t wkbsize)
 {
 	/*
-	byte:     endianness (1 = NDR, 0 = XDR)
+	byte:	  endianness (1 = NDR, 0 = XDR)
 	uint32:   pcid (key to POINTCLOUD_SCHEMAS)
 	uint32:   compression (0 = no compression, 1 = dimensional, 2 = GHT)
 	uchar[]:  data (interpret relative to pcid and compression)
@@ -378,7 +378,7 @@ uint8_t *
 pc_patch_to_wkb(const PCPATCH *patch, size_t *wkbsize)
 {
 	/*
-	byte:     endianness (1 = NDR, 0 = XDR)
+	byte:	  endianness (1 = NDR, 0 = XDR)
 	uint32:   pcid (key to POINTCLOUD_SCHEMAS)
 	uint32:   compression (0 = no compression, 1 = dimensional, 2 = GHT)
 	uchar[]:  data (interpret relative to pcid and compression)
@@ -520,6 +520,68 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
 	}
 
 	return (PCPATCH*)paout;
+}
+
+// first: the first element to select (1-based indexing)
+// count: the number of points to select
+PCPATCH *
+pc_patch_range(const PCPATCH *pa, int first, int count)
+{
+	PCPATCH_UNCOMPRESSED *paout, *pu;
+	int countmax;
+	uint8_t *buf;
+	size_t size;
+	size_t start;
+
+	assert(pa);
+
+	first--;
+	countmax = pa->npoints - first;
+
+	if ( count > countmax )
+		count = countmax;
+
+	if ( first < 0 || count <= 0 )
+		return NULL;
+
+	if ( count == pa->npoints )
+		return (PCPATCH *) pa;
+
+	paout = pc_patch_uncompressed_make(pa->schema, count);
+	if ( !paout )
+		return NULL;
+	paout->npoints = count;
+
+	pu = (PCPATCH_UNCOMPRESSED *) pc_patch_uncompress(pa);
+	if ( !pu )
+	{
+		pc_patch_free((PCPATCH *) paout);
+		return NULL;
+	}
+
+	buf = paout->data;
+	start = pa->schema->size * first;
+	size = pa->schema->size * count;
+
+	memcpy(buf, pu->data + start, size);
+
+	if ( ((PCPATCH *) pu) != pa )
+		pc_patch_free((PCPATCH *) pu);
+
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_extent(paout) )
+	{
+		pcerror("%s: extent computation failed", __func__);
+		pc_patch_free((PCPATCH *) paout);
+		return NULL;
+	}
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_stats(paout) )
+	{
+		pcerror("%s: stats computation failed", __func__);
+		pc_patch_free((PCPATCH *) paout);
+		return NULL;
+	}
+
+	return (PCPATCH *) paout;
 }
 
 /** get point n from patch */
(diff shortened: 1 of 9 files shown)


lib/pc_patch.c
```diff
@@ -522,6 +522,77 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
...
+	PCPATCH_UNCOMPRESSED *paout;
+	uint8_t *buf;
+	size_t size;
+	size_t start;
+
+	assert(pa);
+
+	if(first+count > pa->npoints)
+	count = pa->npoints - first;
+
```
mbredif: Return NULL if count<=0?

lib/pc_patch.c
```diff
@@ -522,6 +522,77 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
...
+	paout = pc_patch_uncompressed_make(pa->schema, count);
+	buf = paout->data;
+	paout->npoints = count;
+	start = pa->schema->size * first;
+	size = pa->schema->size * count;
+
+	switch( pa->type )
```
mbredif: This switch is somewhat redundant with pc_patch_uncompress, which could be used instead (freeing the uncompressed patch if its pointer differs from the input patch.)

lib/pc_patch.c
elemoine: +1

But I am not sure what we will be the result for the user. But I'll try.

elemoine:
> The doc should say if the index is 0 or 1 based. (PointN is currently 1 based but could be made 0 based or vice versa for homogeneity)

Good point. I am going to make it 1-based, just like `PC_PointN` (and PostGIS' `ST_GeometryN`).

elemoine:
> How about adding negative indices for start and count that would count backwards?

Why not, but we don't have that use-case for now. For your information, we use `PC_Range` together with [Revert Morton](https://github.com/Oslandia/pgmorton) in LOPoCS.

lib/pc_patch.c
elemoine: This is again a very good comment. I reimplemented `pc_patch_range` in terms of `pc_patch_umcompress`. Please look at the patch again.

lib/pc_patch.c
elemoine: The result for the user is NULL values. For example:

```
pointcloud=# select id, (pc_range(pa, 1, -1) IS NULL) AS null from patches;
 id | null
----+------
  1 | t
  2 | t
  3 | t
  4 | t
  5 | t
  6 | t
  7 | t
  8 | t
  9 | t
 10 | t
 11 | t
(11 rows)
```

elemoine:
I took into account all the comments from @mbredif. This is ready for review again.

mbredif:
That looks good, thanks elemoine. Maybe some test could be added for edge cases (where NULL patch is returned).

elemoine:
> Maybe some test could be added for edge cases (where NULL patch is returned).

Done.

lib/pc_patch.c
```diff
@@ -522,6 +522,56 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
...
+	size = pa->schema->size * count;
+
+	memcpy(buf, pu->data + start, size);
+
+	if ( ((PCPATCH *) pu) != pa )
+		pc_patch_free((PCPATCH *) pu);
+
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_extent(paout) )
+	{
+		pcerror("%s: extent computation failed", __func__);
```
pblottiere: I think we should free `paout` in this case?

lib/pc_patch.c
```diff
@@ -522,6 +522,56 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
...
+		return NULL;
+	}
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_stats(paout) )
+	{
+		pcerror("%s: stats computation failed", __func__);
```
pblottiere: Same comment about freeing aout`.

pblottiere:
@elemoine Thank you for your work! Aside from my minor comments above, it looks good to me!

lib/pc_patch.c
elemoine: Good catch. And there also is a missing `pc_patch_free(paout)` for the case where `pu == NULL` above. I am fixing those.

elemoine: Done.

lib/pc_patch.c
elemoine: Done.

elemoine:
Thanks @pblottiere and @mbredif. Patch updated. I'll merge it tonight or tomorrow if there are no other comments.

pgsql/pc_access.c
```diff
@@ -677,6 +678,27 @@ Datum pcpatch_pointn(PG_FUNCTION_ARGS)
...
+	PCPATCH *patchout = NULL;
+	if ( patch )
+	{
+		patchout = pc_patch_range(patch, first, count);
+		pc_patch_free(patch);
+	}
+	if ( !patchout )
+		PG_RETURN_NULL();
+	serpaout = pc_patch_serialize(patchout, NULL);
+	PG_RETURN_POINTER(serpaout);
```
mbredif: `patchout` should be freed as well, I guess.

lib/pc_patch.c
```diff
@@ -522,6 +522,61 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
...
+	size_t start;
+
+	assert(pa);
+
+	if ( count < 1 || first < 1 || first > pa->npoints )
+		return NULL;
+
+	if ( first + count > pa->npoints + 1)
+		count = pa->npoints - first + 1;
+
```
mbredif: `if ( count == pa->npoints ) return pa;` ? (implies a conditionnal free [here](https://github.com/pgpointcloud/pointcloud/pull/152/files#diff-553ae5a53a7958f15eea808d2b303e82R694) )

lib/pc_patch.c
```diff
@@ -522,6 +522,61 @@ pc_patch_from_patchlist(PCPATCH **palist, int numpatches)
...
+PCPATCH *
+pc_patch_range(const PCPATCH *pa, int first, int count)
+{
+	PCPATCH_UNCOMPRESSED *paout, *pu;
+	uint8_t *buf;
+	size_t size;
+	size_t start;
+
+	assert(pa);
+
```
mbredif: You could do first `--first;` up front to go from 1-based to 0-based indexing in order to simplify the indexing : no `-1` line 566 and 537-541 could become :

```
  --first;
  int countmax = pa->npoints - first;
  if ( count > countmax ) count = countmax ;
  if ( first < 0 || count <= 0) return NULL;
```

lib/pc_patch.c
elemoine: Do you mean `if ( first == 0 && count == pa->npoints ) return pa;` instead?

elemoine: Ok, `count == pa->npoints` implies that `first == 0`.

pgsql/pc_access.c
elemoine: Done.

lib/pc_patch.c
elemoine: Done. I like that a lot!

lib/pc_patch.c
elemoine: Done.

elemoine:
Thanks for the excellent comments @mbredif. I took all of them into account.

lib/pc_patch.c
elemoine: And I added `test_patch_range_compression_none_with_full_range` for that case.

PR #153, opened 2017-03-23
Use eclint 1.1.5

elemoine:
eclint 2 was recently released. Unfortunately this version is not compatible with the version of NodeJS available on Travis. So let's stick to eclint 1.1.5 for now.


pblottiere:
@elemoine It seems good to me.

Thank you.

PR #157, opened 2017-03-29
Add an SQL Tests section to README.md

elemoine:
I tend to forget what I need to do for running the SQL Tests (yes, I know…). Hence this addition to the README file…


elemoine:
> LGTM. Maybe sudo make install instead ?

Yes, `sudo` is specified elsewhere in the README so it makes sense to use it here as well.

Done.

PR #158, opened 2017-03-29
Add PC_BoundingDiagonal(p PCPATCH)

elemoine:
This provides a `PC_BoundingDiagonal` function that can be used to get the 2D, Z, M or ZM bounding diagonal of a patch. Using this function indices on `PC_BoundingDiagonal(patch)` can be created, making 3D spatial relationship tests possible.

@mbredif @pblottiere does this correspond to what we discussed?

Todo:

* [x] Document `PC_BoundingDiagonalAsBinary`
* [x] Add SQL tests
* [x] Add unit tests

### DIFF
--- a/lib/pc_api.h
+++ b/lib/pc_api.h
@@ -431,6 +431,9 @@ int pc_bytes_deserialize(const uint8_t *buf, const PCDIMENSION *dim, PCBYTES *pc
 /** Wrap serialized stats in a new stats objects */
 PCSTATS* pc_stats_new_from_data(const PCSCHEMA *schema, const uint8_t *mindata, const uint8_t *maxdata, const uint8_t *avgdata);
 
+/** Allocate a stats object */
+PCSTATS* pc_stats_new(const PCSCHEMA *schema);
+
 /** Free a stats object */
 void pc_stats_free(PCSTATS *stats);
 
@@ -445,6 +448,12 @@ int pc_patch_compute_extent(PCPATCH *patch);
 
 /** True/false if bounds intersect */
 int pc_bounds_intersects(const PCBOUNDS *b1, const PCBOUNDS *b2);
+
+/** Returns OGC WKB of the bounding diagonal of XY bounds */
+uint8_t* pc_bounding_diagonal_wkb_from_bounds(const PCBOUNDS *bounds, const PCSCHEMA *schema, size_t *wkbsize);
+
+/** Returns OGC WKB of the bounding diagonal of XY, XYZ, XYM or XYZM bounds */
+uint8_t* pc_bounding_diagonal_wkb_from_stats(const PCSTATS *stats, size_t *wkbsize);
 
 /** Subset batch based on less-than condition on dimension */
 PCPATCH* pc_patch_filter_lt_by_name(const PCPATCH *pa, const char *name, double val);
--- a/lib/pc_util.c
+++ b/lib/pc_util.c
@@ -156,6 +156,30 @@ wkb_get_int16(const uint8_t *wkb, int flip_endian)
 		return i;
 }
 
+uint8_t *
+wkb_set_double(uint8_t *wkb, double d)
+{
+	memcpy(wkb, &d, 8);
+	wkb += 8;
+	return wkb;
+}
+
+uint8_t *
+wkb_set_uint32(uint8_t *wkb, uint32_t i)
+{
+	memcpy(wkb, &i, 4);
+	wkb += 4;
+	return wkb;
+}
+
+uint8_t *
+wkb_set_char(uint8_t *wkb, char c)
+{
+	memcpy(wkb, &c, 1);
+	wkb += 1;
+	return wkb;
+}
+
 uint32_t
 wkb_get_pcid(const uint8_t *wkb)
 {
@@ -206,6 +230,7 @@ wkb_get_npoints(const uint8_t *wkb)
 	}
 	return npoints;
 }
+
 uint8_t*
 uncompressed_bytes_flip_endian(const uint8_t *bytebuf, const PCSCHEMA *schema, uint32_t npoints)
 {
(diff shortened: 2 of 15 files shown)


pgsql/pc_inout.c
```diff
@@ -313,6 +314,37 @@ Datum pcpatch_bytea_envelope(PG_FUNCTION_ARGS)
...
+{
+	SERIALIZED_PATCH *serpatch;
+	PCSCHEMA *schema;
+	PCPATCH *patch;
+	uint8 *bytes;
+	size_t bytes_size;
+	bytea *wkb;
+	size_t wkb_size;
+
+	serpatch = PG_GETARG_SERPATCH_P(0);
```
mbredif: You should not deserialize the whole patch. See the stats access functions for deserializing only the stats.

pgsql/pc_inout.c
```diff
@@ -32,6 +32,7 @@ Datum pcpoint_as_text(PG_FUNCTION_ARGS);
 Datum pcpatch_as_text(PG_FUNCTION_ARGS);
 Datum pcpoint_as_bytea(PG_FUNCTION_ARGS);
 Datum pcpatch_bytea_envelope(PG_FUNCTION_ARGS);
+Datum pcpatch_bytea_bounding_diagonal(PG_FUNCTION_ARGS);
```
mbredif: I would prefer `pcpatch_bounding_diagonal_as_bytea`. (I propose to rename the `pcpatch_bytea_envelope` to `pcpatch_envelope_as_bytea` in my pc_envelope PR.

mbredif:
> Document PC_BoundingEnvelopeAsBinary

I assume you meant `PC_BoundingDiagonalAsBinary`.

pgsql/pc_pgsql.h
```diff
@@ -116,6 +116,9 @@ char* pc_patch_to_hexwkb(const PCPATCH *patch);
 /** Returns OGC WKB for envelope of PCPATCH */
 uint8_t* pc_patch_to_geometry_wkb_envelope(const SERIALIZED_PATCH *pa, const PCSCHEMA *schema, size_t *wkbsize);
 
+/** Returns OGC WKB for bounding diagonal of PCPATCH */
+uint8_t* pc_patch_to_geometry_wkb_bounding_diagonal(const PCPATCH *pa, const PCSCHEMA *schema, size_t *wkbsize);
+
```
mbredif: I would rather have this function (and  `pc_patch_to_geometry_wkb_envelope`, see [this commit](https://github.com/LI3DS/pointcloud/pull/59/commits/788dc496ab536959997784bed37d495e3154d21d)) in the lib, so that it can be tested with make check and valgrind more easily.

pgsql/pc_inout.c
elemoine: +1

pgsql/pc_inout.c
elemoine: Yes, this was my intent.

pgsql/pc_pgsql.h
elemoine: Moving the function to `lib` was also my intent.

pgsql/pc_inout.c
elemoine: Done.

pgsql/pc_pgsql.h
elemoine: Done.

pgsql/pc_inout.c
mbredif: Not really, see https://github.com/pgpointcloud/pointcloud/blob/master/pgsql/pc_access.c#L711-L725

lib/pc_util.c
```diff
@@ -263,3 +288,80 @@ void pc_bounds_merge(PCBOUNDS *b1, const PCBOUNDS *b2)
 	if ( b2->ymax > b1->ymax ) b1->ymax = b2->ymax;
 }
 
+uint8_t *
+pc_bounds_xyzm_to_bounding_diagonal_wkb(
+	const PCSTATS *stats, const PCBOUNDS *bounds, const PCSCHEMA *schema,
+	size_t *wkbsize)
```
mbredif: it is a little cumbersome to pass both the stats and the bounds.
Is it much more efficient to use the bounds rather than the stats only ?
schema is already accessible in the points of stats.

elemoine: Agree. How about having two functions, `pc_bounds_to_bounding_diagonal_wkb` and `pc_bounds_xyzm_bounding_diagonal_wkb`, and use the former as an optimization for the 2D case? See the updated patch.

elemoine:
> Not really, see https://github.com/pgpointcloud/pointcloud/blob/master/pgsql/pc_access.c#L711-L725

Done. Thank you for pointing me to this technique.

lib/pc_util.c
```diff
@@ -263,3 +288,123 @@ void pc_bounds_merge(PCBOUNDS *b1, const PCBOUNDS *b2)
 	if ( b2->ymax > b1->ymax ) b1->ymax = b2->ymax;
 }
 
+static uint32_t srid_mask = 0x20000000;
+static uint32_t m_mask = 0x40000000;
+static uint32_t z_mask = 0x80000000;
+
+uint8_t *
+pc_bounds_to_bounding_diagonal_wkb(
```
elemoine: Could be changed to `pc_bounding_diagonal_wkb_from_bounds`. And the second function could be changed to `pc_bounding_diagonal_wkb_from_stats`.

lib/pc_util.c
```diff
@@ -263,3 +288,123 @@ void pc_bounds_merge(PCBOUNDS *b1, const PCBOUNDS *b2)
...
+	if ( wkbsize )
+		*wkbsize = size;
+
+	return wkb;
+}
+
+uint8_t *
+pc_bounds_xyzm_to_bounding_diagonal_wkb(
+	const PCSTATS *stats, const PCSCHEMA *schema, size_t *wkbsize)
+{
```
mbredif: Why not simply `pc_stats_to_bounding_diagonal_wkb(const PCSTATS *stats, size_t *wkbsize)` ?
`schema` could be retrieved as `stats->min->schema`

lib/pc_api.h
```diff
@@ -437,6 +440,12 @@ int pc_patch_compute_extent(PCPATCH *patch);
 /** True/false if bounds intersect */
 int pc_bounds_intersects(const PCBOUNDS *b1, const PCBOUNDS *b2);
 
+/** Returns OGC WKB of the bounding diagonal of XY bounds */
+uint8_t* pc_bounds_to_bounding_diagonal_wkb(const PCBOUNDS *bounds, const PCSCHEMA *schema, size_t *wkbsize);
+
+/** Returns OGC WKB of the bounding diagonal of XYZ, XYM or XYZM bounds */
```
mbredif: XY is ok too

lib/pc_util.c
mbredif: That is also better than the current proposal with `bounds_xyzm`...

elemoine: Done.

lib/pc_util.c
elemoine: Indeed. Done.

lib/pc_api.h
elemoine: Done.

elemoine:
So with this patch we can create this index on a patch table:

```sql
create index on my_table using gist(pc_boundingdiagonal(patch)  gist_geometry_ops_nd);
```

and then use index-based queries like this:

```sql
select patch from my_table where pc_boundingdiagonal(patch) &&& 'linestring z (…)'::geometry;
```

I did some preliminary tests with our LOPoCS visualization server and it looks promising!

pgsql/pc_inout.c
```diff
@@ -313,6 +314,50 @@ Datum pcpatch_bytea_envelope(PG_FUNCTION_ARGS)
...
+		bytes = pc_bounding_diagonal_wkb_from_bounds(
+			&serpatch->bounds, schema, &bytes_size);
+	}
+
+	wkb_size = VARHDRSZ + bytes_size;
+	wkb = palloc(wkb_size);
+	memcpy(VARDATA(wkb), bytes, bytes_size);
+	SET_VARSIZE(wkb, wkb_size);
+
+	pfree(bytes);
```
mbredif: `pcfree` ?

mbredif:
Good work @elemoine. I think this can be merged after rebasing on the merge of #160 and #149 .

pgsql/pc_inout.c
elemoine: We use `pfree` (from Postgres) elsewhere for the `pgsql` code.

mbredif: My remark is that `bytes` is allocated with `pcalloc`, so I would expect freeing it with `pcfree`. I do not know if it matters, though.

elemoine:
I rebased this, and updates the NEWS file to mention the two new functions.

pgsql/pc_inout.c
elemoine: `pcpatch_envelope_as_bytea` uses `pfree`, but you're right that it makes more sense to use `pcfree` in that case. I'll change it.

elemoine: Done.

elemoine:
@pblottiere @strk This patch is ready. @mbredif reviewed it, and I took all his comments into account. Feel free to review this and provide comments. Otherwise I'll merge it in the coming days. Thanks.

lib/pc_util.c
```diff
@@ -156,6 +156,30 @@ wkb_get_int16(const uint8_t *wkb, int flip_endian)
 		return i;
 }
 
+uint8_t *
+wkb_set_double(uint8_t *wkb, double d)
+{
+	memcpy(wkb, &d, 8);
+	wkb += 8;
```
pblottiere: A minor comment: we could use `sizeof` instead of having it hardcoded.

lib/pc_util.c
```diff
@@ -156,6 +156,30 @@ wkb_get_int16(const uint8_t *wkb, int flip_endian)
...
+	return wkb;
+}
+
+uint8_t *
+wkb_set_uint32(uint8_t *wkb, uint32_t i)
+{
+	memcpy(wkb, &i, 4);
+	wkb += 4;
```
pblottiere: Same comment about `sizeof`

lib/pc_util.c
```diff
@@ -156,6 +156,30 @@ wkb_get_int16(const uint8_t *wkb, int flip_endian)
...
+	return wkb;
+}
+
+uint8_t *
+wkb_set_char(uint8_t *wkb, char c)
+{
+	memcpy(wkb, &c, 1);
+	wkb += 1;
```
pblottiere: Same comment about `sizeof`

pblottiere:
@elemoine Very nice work! It's gonna be be very useful to LOPoCS!

lib/pc_util.c
elemoine: I agree that it's better to use `sizeof` but the `wkb_get_*` functions and other functions in this file use numerical values instead of `sizeof`. I'd rather favor consistency and use numerical values here in the setters as well.

PR #143, opened 2017-02-20
testing : pcpoint_get/set and pc_point_to_geometry_wkb

mbredif:
This PR :
- adds simple xy, xyz, xym and xyzm xml schemas for unit tests
- defines CU_ASSERT_FAILURE and CU_ASSERT_SUCCESS for easy checking of return values in tests
- provides a consistent API for pcpoint_get/set functions with tests (`test_point_xyzm`). (e.g. pcpoint_set functions were returning the PC_SUCCESS/PC_FAILURE as a double !)
- provides tests for pcpoint to geometry casts (`test_point_geometry`)
- replaces `int32_t x_position` with `PCDIMENSION *xdim` (and similarly for y,z,m) in PCSCHEMA to get a direct access to the corresponding dimension (code readability improvement by discarding the `-1` constant and slight expected performance increase of get/set functions)

PR #160, opened 2017-03-30
Point xyzm test

mbredif:
#143 with fixed warings


mbredif:
Rebased, thanks @elemoine for pointing to `git rebase --onto`.

PR #161, opened 2017-04-02
Compile with -Werror on Travis

elemoine:
This PR adds `-Werror` to `CFLAGS` on Travis. This is to avoid merging code that introduces new compile warnings.


pblottiere:
@elemoine Good idea!

Issue #93, opened 2015-10-18
pointcloud_abs_in and pointcloud_abs_out should not be volatile


robe2:
I have to double check if I get this on lower than 9.5, so this might be a new thing  in 9.5,

When I do 

```

CREATE EXTENSION pointcloud;

```

I get these warnings:

```
WARNING:  type input function pointcloud_abs_in should not be volatile
WARNING:  type output function pointcloud_abs_out should not be volatile

```

strk:
That's one of those things you can't easily upgrade.
But yes, it would be nice to fix for new installs.

DBlasby:
I found this issue, too.  Is it something to ignore or is a problem?

PR #162, opened 2017-04-03
Make pointcloud_abs_out|in immutable and strict

elemoine:
Type I/O functions should not be volatile. PostgreSQL 9.5 and above include a test for this, yielding this warning when a volatile function is detected:

    WARNING:  type input function pointcloud_abs_in should not be volatile

This commit makes pointcloud_abs_out and pointcloud_abs_in, which are currently empty shells, non-volatile.

Fixes https://github.com/pgpointcloud/pointcloud/issues/93.


elemoine:
@strk do you agree with that patch? I guess we don't need to make this upgradable, as I believe the warnings are harmless.

strk:
Yes it sounds fine by me.

elemoine:
> Yes it sounds fine by me.

Thanks. I'll merge that if I don't get negative feedback.

PR #164, opened 2017-04-14
Add the geometry(pcpach) function back

elemoine:
For backward-compatibility reasons this commit adds back the `geometry(pcpatch)` function that was removed with 7da35f0. This function is just an alias to `PC_Envelope(pcpatch)`.

### DIFF
--- a/pgsql_postgis/pointcloud_postgis--1.0.sql
+++ b/pgsql_postgis/pointcloud_postgis--1.0.sql
@@ -23,7 +23,15 @@ CREATE OR REPLACE FUNCTION PC_Envelope(pcpatch)
 	$$
 	LANGUAGE 'sql';
 
+CREATE OR REPLACE FUNCTION geometry(pcpatch)
+	RETURNS geometry AS
+	$$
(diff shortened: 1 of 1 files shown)


elemoine:
@mbredif @pblottiere what do you think?

mbredif:
I suggest to keep this PR open for a while and eventually merge it if someone complains.

elemoine:
> Thus, I suggest to keep this PR open for a while and eventually merge it if someone complains too loudly.

I don't know who uses pointcloud right now, but maybe people use releases instead of building master from source, meaning that it would be a bit too late when someone complains. My suggestion is to maintain backward-compatibility, The `geometry` cast for `PCPATCH` is not documented, so I think it's fine to leave keep it for people who already rely on it.

tomvantilburg:
For me it would have saved some work to have it, for people running in production it seems needed. Also, I appreciate the idea of keeping deprecated functions hanging around for a while (incuding an invasive deprecation warning ;-).
So +1 on bringing a deprecated Geometry(pcpatch) back.

PR #165, opened 2017-04-14
Add PC_Transform

elemoine:
This adds a `PC_Transform` function to transform a patch based on new schema.

This is still WIP at the moment.

### DIFF
--- a/lib/pc_patch.c
+++ b/lib/pc_patch.c
@@ -724,3 +724,86 @@ pc_patch_set_schema(PCPATCH *patch, const PCSCHEMA *new_schema, double def)
 
 	return (PCPATCH*) paout;
 }
+
+
+/**
+* Read all the points from "patch", and transform them based on "new_schema".
+* Return a new patch with the transformed points.
+*/
+PCPATCH*
+pc_patch_transform(const PCPATCH *patch, const PCSCHEMA *new_schema, double def)
+{
+	PCDIMENSION** new_dimensions = new_schema->dims;
+	PCDIMENSION* old_dimensions[new_schema->ndims];
+	const PCSCHEMA *old_schema = patch->schema;
+	PCPATCH_UNCOMPRESSED *paout;
+	PCPOINT opt, npt;
+	PCPATCH *pain;
+	size_t i, j;
+
+	if ( old_schema->srid != new_schema->srid )
+	{
+		pcwarn("old and new schemas have different srids, and data "
+			   "reprojection is not yet supported");
+		return NULL;
+	}
+
+	for ( j = 0; j < new_schema->ndims; j++ )
+	{
+		PCDIMENSION *ndim = new_dimensions[j];
+		PCDIMENSION *odim = pc_schema_get_dimension_by_name(
+				old_schema, ndim->name);
+		old_dimensions[j] = odim;
+	}
+
+	pain = pc_patch_uncompress(patch);
+
+	paout = pc_patch_uncompressed_make(new_schema, patch->npoints);
+	paout->npoints = pain->npoints;
+
+	opt.schema = old_schema;
+	npt.schema = new_schema;
+	opt.readonly = PC_TRUE;
+	npt.readonly = PC_TRUE;
+
+	opt.data = ((PCPATCH_UNCOMPRESSED *) pain)->data;
+	npt.data = paout->data;
+
+	// reinterpret the data and fill the output patch
+	//
+	// TODO: for the case where the old and new dimension sets don't intersect (all
+	// the values in old_dimensions are NULL) a faster path could probably be used
+	for ( i = 0; i <patch->npoints; i++ )
+	{
+		for ( j = 0; j < new_schema->ndims; j++ )
+		{
+			// pc_point_get_double returns immediately w/o changing val if the
+			// dimension it is passed is NULL
+			double val = def;
+			pc_point_get_double(&opt, old_dimensions[j], &val);
+			pc_point_set_double(&npt, new_dimensions[j], val);
+		}
+
+		opt.data += old_schema->size;
+		npt.data += new_schema->size;
+	}
+
+	if ( pain != patch )
+		pc_patch_free(pain);
+
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_extent(paout) )
+	{
+		pcerror("%s: failed to compute patch extent", __func__);
+		pc_patch_free((PCPATCH *)paout);
+		return NULL;
+	}
+
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_stats(paout) )
+	{
+		pcerror("%s: failed to compute patch stats", __func__);
+		pc_patch_free((PCPATCH *)paout);
+		return NULL;
+	}
+
+	return (PCPATCH*) paout;
+}
--- a/pgsql/pc_editor.c
+++ b/pgsql/pc_editor.c
@@ -10,85 +10,117 @@
 #include "pc_pgsql.h"	   /* Common PgSQL support for our type */
 
 Datum pcpatch_setpcid(PG_FUNCTION_ARGS);
+Datum pcpatch_transform(PG_FUNCTION_ARGS);
 
 
-static bool
-pcpatch_schema_same_dimensions(const PCSCHEMA *s1, const PCSCHEMA *s2)
+static SERIALIZED_PATCH *
+pcpatch_set_schema(SERIALIZED_PATCH *serpa,
+				  PCSCHEMA *oschema, PCSCHEMA *nschema, float8 def)
 {
-	size_t i;
-
-	if ( s1->ndims != s2->ndims )
-		return false;
-
-	for ( i = 0; i < s1->ndims; i++ )
-	{
-		PCDIMENSION *s1dim = s1->dims[i];
-		PCDIMENSION *s2dim = s2->dims[i];
-
-		if ( strcasecmp(s1dim->name, s2dim->name) != 0 )
-			return false;
-
-		if ( s1dim->interpretation != s2dim->interpretation )
-			return false;
-	}
-
-	return true;
-}
-
-
-PG_FUNCTION_INFO_V1(pcpatch_setpcid);
-Datum pcpatch_setpcid(PG_FUNCTION_ARGS)
-{
-	PCPATCH *paout = NULL;
 	SERIALIZED_PATCH *serpatch;
-	SERIALIZED_PATCH *serpa = PG_GETARG_SERPATCH_P(0);
-	int32 pcid = PG_GETARG_INT32(1);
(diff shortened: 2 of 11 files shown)


elemoine:
@mbredif what do you think about this? Does that also match what you had in mind?

mbredif:
Yes indeed! My comments are :
- similar handling of pcpoints in `pc_patch_transform` as in `pc_patch_set_schema`
- Intercept PC_Transform calls that would have obviously yielded the same result with PC_SetPCId and apply `pc_patch_set_schema` instead (same pcid, same scale/offset...)
- PROJ4 support as a cmake option (-> another PR ?)

elemoine:
> similar handling of pcpoints in pc_patch_transform as in pc_patch_set_schema

Can you please be more specific?

> Intercept PC_Transform calls that would have obviously yielded the same result with PC_SetPCId and apply pc_patch_set_schema instead (same pcid, same scale/offset...)

I thought that the user should use `PC_SetPCId` for the cases where `PC_SetPCId` can be used. But I can re-consider this.


> PROJ4 support as a cmake option (-> another PR ?)

Yes, a separate PR was my plan.

mbredif:
>>    similar handling of pcpoints in pc_patch_transform as in pc_patch_set_schema

> Can you please be more specific?

`pc_patch_set_schema` initializes points on the stack pointing to uncompressed input and output patches, whereas `pc_patch_transform` relies on pointlists of dynamically allocated pcpoints and thus needs extra pcalloc/pcfree/memcpy calls.

> I thought that the user should use PC_SetPCId for the cases where PC_SetPCId can be used. But I can re-consider this.

I do not know what is the usual approach in postgis or postgres in general. I suggest to stick with that one (with a note in the readme to document whether this optimisation is automatic or not)

elemoine:
> pc_patch_set_schema initializes points on the stack pointing to uncompressed input and output patches, whereas pc_patch_transform relies on pointlists of dynamically allocated pcpoints and thus needs extra pcalloc/pcfree/memcpy calls.

Good call!

> I do not know what is the usual approach in postgis or postgres in general. I suggest to stick with that one (with a note in the readme to document whether this optimisation is automatic or not)

I added a commit that make pcpatch_transform use setpcid as a fast path when possible.

lib/pc_patch.c
```diff
@@ -724,3 +724,86 @@ pc_patch_set_schema(PCPATCH *patch, const PCSCHEMA *new_schema, double def)
...
+		npt.data += new_schema->size;
+	}
+
+	if ( pain != patch )
+		pc_patch_free(pain);
+
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_extent(paout) )
+	{
+		pcerror("%s: failed to compute patch extent", __func__);
+		return NULL;
```
mbredif: `pc_patch_free(paout);` ?

lib/pc_patch.c
```diff
@@ -724,3 +724,86 @@ pc_patch_set_schema(PCPATCH *patch, const PCSCHEMA *new_schema, double def)
...
+	}
+
+	if ( PC_FAILURE == pc_patch_uncompressed_compute_stats(paout) )
+	{
+		pcerror("%s: failed to compute patch stats", __func__);
+		return NULL;
```
mbredif: `pc_patch_free(paout);` ?

pgsql/pc_editor.c
```diff
@@ -10,85 +10,117 @@
...
 		{
 			// no need to deserialize the patch
 			serpatch = palloc(serpa->size);
 			if ( ! serpatch )
-				PG_RETURN_NULL();
+				return NULL;
 			memcpy(serpatch, serpa, serpa->size);
-			serpatch->pcid = pcid;
-			PG_RETURN_POINTER(serpatch);
+			serpatch->pcid = serpa->pcid;
```
mbredif: I think you meant `serpatch->pcid = nschema->pcid;`.
do you have a test case for that ?

pgsql/pc_editor.c
```diff
@@ -10,85 +10,117 @@
...
 
 Datum pcpatch_setpcid(PG_FUNCTION_ARGS);
+Datum pcpatch_transform(PG_FUNCTION_ARGS);
 
 
-static bool
-pcpatch_schema_same_dimensions(const PCSCHEMA *s1, const PCSCHEMA *s2)
+static SERIALIZED_PATCH *
+pcpatch_set_schema(SERIALIZED_PATCH *serpa,
+				  PCSCHEMA *oschema, PCSCHEMA *nschema, float8 def)
```
mbredif: Indenting is weird here.

lib/pc_patch.c
```diff
@@ -724,3 +724,86 @@ pc_patch_set_schema(PCPATCH *patch, const PCSCHEMA *new_schema, double def)
...
+			PCDIMENSION *ndim = new_dimensions[j];
+			PCDIMENSION *odim = old_dimensions[j];
+
+			// pc_point_get_double returns immediately w/o changing val if the
+			// dimension is NULL
+			val = def;
+			pc_point_get_double(&opt, odim, &val);
+
+			pc_point_set_double(&npt, ndim, val);
+		}
```
mbredif: This could be condensed as (but that is a matter of taste): 
```
// pc_point_get_double returns immediately w/o changing val if the
// dimension is NULL
double val = def;
pc_point_get_double(&opt, old_dimensions[j], &val);
pc_point_set_double(&npt, new_dimensions[j], val);
```

pgsql/pc_editor.c
elemoine: It looks weird on GitHub, but it isn't in reality.

pgsql/pc_editor.c
elemoine: Good catch. Yes, that's covered: https://github.com/LI3DS/pointcloud/blob/36a2fc52284be39858263cb4e2ef4330ca3629d0/pgsql/sql/pointcloud.sql#L417-L422.

lib/pc_patch.c
elemoine: Good catch!

lib/pc_patch.c
elemoine: Thanks.

elemoine:
@mbredif I took all your comments into account, and I added unit tests for `pc_schema_same_dimensions` and `pc_schema_same_interpretations`. I also want to rewrite the Git history before merging.

lib/pc_patch.c
elemoine: Yep, I like the condensed version better as well.

elemoine:
> I also want to rewrite the Git history before merging.

Done. It's now only one commit.

lib/pc_patch.c
```diff
@@ -724,3 +724,83 @@ pc_patch_set_schema(PCPATCH *patch, const PCSCHEMA *new_schema, double def)
...
+	}
+
+	for ( j = 0; j < new_schema->ndims; j++ )
+	{
+		PCDIMENSION *ndim = new_dimensions[j];
+		PCDIMENSION *odim = pc_schema_get_dimension_by_name(
+				old_schema, ndim->name);
+		old_dimensions[j] = odim;
+	}
+
```
pblottiere: Maybe we could add a warning if no dimensions are found in common with `old_shema` (i.e when `old_dimensions` only contains null elements)?

pblottiere:
Apart from my brief remark, it looks good to me! Very nice work!

lib/pc_patch.c
elemoine: I don't know. I tend to think that this should be handled as a normal condition and that no message should be output.

@mbredif any opinion about this?

pblottiere: I'm not sure it's a normal condition. Who would like to transform something which has no dimensions in common? It would likely be an error from the user.

But I don't have a strong opinion here.

elemoine: Yep, I understand. But let's say you need to transform a bunch of different patches, some of them with a set of dimensions that intersects the new set, and others with a set dimensions that do not intersect the new set. In that case I am not sure we want to display a log message for each patch of the second sort. This may pollute the output, with no value for the user, who knows that there are patches with dimensions that do not intersect the new set.

I have no strong opinion either. Just asking this question to myself.

pblottiere: > This may pollute the output, with no value for the user, who knows that there are patches with dimensions that do not intersect the new set.

Indeed...

mbredif: @pblottiere, what would you do ? hint to use `PC_MakePoint(pcid integer, vals float8[])` instead ?

For now X and Y dimensions are always present so it should never be the case, unless different aliases are used (eg: [X, Y] vs [Lon, Lat], as pc_transform is not case sensitive, but the Y dimension could eg be named "Y", "Latitude" or "Lat")

I think I would leave it like that.
If we decide it to be a normal condition, how difficult would it be to automatically switch to an automatically even faster path ? (no deserialization and no interpretation). Would we gain much in terms of speed ?

elemoine: > If we decide it to be a normal condition, how difficult would it be to automatically switch to an automatically even faster path ? (no deserialization and no interpretation). Would we gain much in terms of speed ?

I don't know, but this sounds like premature optimization to me, for a case that might not happen in real life.

pblottiere: > what would you do ? 

I was thinking about stopping the `pc_transform` function with a warning message as soon as this condition is reached.

But, as you said, if we consider this case as a normal condition (no warning message), then we should switch to a faster path instead of looping over each points for each dimensions into the new schema.

pblottiere: > this sounds like premature optimization 

It also adds a touch of elegance. But, again, I don't have a string opinion here.

mbredif: I tend to favor to leave like this, merge it and open an issue on the topic for further discussion on whether it should rather fail, continue with the current implementation, or continue with an optimized implementation.

pblottiere: > I tend to favor to leave like this

It's OK for me. Sorry for the noise!

elemoine: > But, as you said, if we consider this case as a normal condition (no warning message), then we should switch to a faster path instead of looping over each points for each dimensions into the new schema.

We still need a loop to create points, and an inner loop to create dimension values for each point. So the current code doesn't sound that bad for this case in my opinion. How would you do it differently?

elemoine: > I tend to favor to leave like this, merge it and open an issue on the topic for further discussion on whether it should rather fail, continue with the current implementation, or continue with an optimized implementation.

Or instead of an issue that will stay there for ever I can add a comment in the code. It may also stay there for a long time, but I think it's a better place for this sort of things.

pblottiere: > Or instead of an issue that will stay there for ever I can add a comment in the code. 

+1. It will avoid  to ask ourselves when reaching this part of the code

PR #166, opened 2017-04-18
Use Artistic Style for code formatting

elemoine:
This PR suggests using [Artistic Style](http://astyle.sourceforge.net) for formatting pointcloud's C code in a consistent way.

We already use EditorConfig, but EditorConfig doesn't know about C, so it is lenient and doesn't enforce C code that is fully consistent.

The astyle config used (in `tools/astyle.sh`) is compatible with our EditorConfig config, so we can keep EditorConfig and just add astyle for making the C code more consistent.

Thoughts?


mbredif:
No strong opinion, if your are sure it does not interfere with editorconfig.
It does mean that, for security, a user would have have to run  `./tools/astyle.sh` before each commit, is that right ?

elemoine:
Yes, and the build will fail there are files that are ill-formatted.

elemoine:
@pblottiere @strk any opinion about this?

strk:
My opinion on code style is: use whatever you want, as long as it's readable.

Enforcing a specific style is ok to me only when:

   1. It's easy for contributors to obtain conformancy

   2. It's easy for maintainers to check for conformancy

elemoine:
> It's easy for contributors to obtain conformancy

Yes, it's a easy. Contributors can just run `./tools/astyle.sh` to make their changes in conformance with our code style.

> It's easy for maintainers to check for conformancy

The conformancy is checked in the Travis process.

strk:
When we tried the astyle route with PostGIS we quickly found out
that different astyle versions resulted in different behaviour.

QGIS fixed that by embedding their own version of the astyle source code.
Is this the plan for pgpointcloud too ?

elemoine:
Good question. I don't know. And I did not know that astyle was so "unstable". On Travis we could rely on a specfic version of astyle  that we build ourselves as part of the Travis build process. But I wouldn't like that we tell contributors that they need to compile astyle to test the style of their code locally.

strk:
Well, they need to build anyway, so they would not really notice that
an additional .c file is being built (yes, it's a single .c file in QGIS).

Compare this to the struggle of a contributor with non-conformant version
of astyle who has to download and build separately a specific version
(will it still be available in 2 years?).

elemoine:
Ok, I did not realize this was just one `.c` file. So embedding our own astyle code may make sense. I'll give it a try.

elemoine:
It is not a single .c file in QGIS as far as I can see: https://github.com/qgis/QGIS/tree/master/src/astyle.

strk:
Oops, I was wrong.

elemoine:
Closing this for now. I may reopen it in the future if I feel brave enough.

PR #167, opened 2017-04-20
Ignore case where comparing dimension names

elemoine:
The case of the characters should be ignored when comparing dimension names in pcpatch_schema_same_dimensions. This is to match the behavior of pc_schema_get_dimension_by_name which ignores the case.


mbredif:
Good catch. I would merge it if I had the rights :)

elemoine:
Thanks for the review.

Issue #171, opened 2017-05-08
pgpointcloud no longer compiles against mingw64


robe2:
I think this has been failing since around January under mingw64.

`gcc -o cu_tester cu_tester.o cu_pc_bytes.o cu_pc_schema.o cu_pc_point.o cu_pc_patch.o cu_pc_patch_ght.o cu_pc_patch_lazperf.o cu_pc_sort.o cu_pc_util.o ../libpc.a ../liblazperf.a -L/projects/libxml/rel-libxml2-2.7.8w64gcc48/lib -lxml2 -lws2_32 -L/projects/zlib/rel-1.2.8w64gcc48/lib -L/projects/libxml/rel-libxml2-2.7.8w64gcc48/lib -lcunit -L/projects/zlib/rel-1.2.8w64gcc48/lib -L/projects/libxml/rel-libxml2-2.7.8w64gcc48/lib -lz  -lm -lstdc++
../libpc.a(pc_sort.o): In function `sort_r':
e:\jenkins\postgis\pointcloud\pointcloud\lib/sort_r/sort_r.h:193: undefined reference to `qsort_r'
collect2.exe: error: ld returned 1 exit status
Makefile:44: recipe for target `cu_tester' failed
make[2]: *** [cu_tester] Error 1
make[2]: Leaving directory `/projects/postgis/pointcloud/pointcloud/lib/cunit'
Makefile:32: recipe for target `all' failed
make[1]: *** [all] Error 2
make[1]: Leaving directory `/projects/postgis/pointcloud/pointcloud/lib'
Makefile:3: recipe for target `all' failed
make: *** [all] Error 2`

The culprit seems to be in sort_r.h because it's being caugh by the __MINGW32__ check.

If I change the:

`defined __MINGW32__`

to 

`(defined __MINGW32__ && !defined __MINGW64__)`

It then compiles okay,

Not sure it messes anyone up to make that change though.
I'll submit a pull request in a bit.

Having trouble having it detect my CUnit as well so need to work that out.

robe2:
Okay I committed a pull request.  I had to go with the __MINGW64_VERSION_MAJOR check since for the 32-bit flavor of mingw-w64, the obvious __MINGW64__ is not defined and it required the same fix.

robe2:
Okay that fixed my compile.

PR #172, opened 2017-05-08
fix compile error for mingw64 (both 64-bit and 32-bit variants). Issu…

robe2:
Fix for issue  #171


mbredif:
LGTM
You may want to have a discussion and file a PR upstream as well:
http://github.com/noporpoise/sort_r
Thanks for the fix!

elemoine:
Based on https://sourceforge.net/p/mingw-w64/wiki2/Answer%20Check%20For%20Mingw-w64/, it looks like the right fix to me.

PR #176, opened 2017-05-11
Fix compilation with libght

elemoine:
This PR fixes compilations errors in the pointcloud ght code. We don't builld with libght enabled on Travis, so we do not always see errors.

Fixes #175.

Issue #175, opened 2017-05-09
pc_patch_ght.c:166:25: error: PCSCHEMA


animecomico:
Hi,
I have this error in build master branch: 
pointcloud/lib/pc_patch_ght.c:166:25: error: ‘PCSCHEMA {aka const struct <anonymous>}’ has no member named ‘x_position’
     if ( j == pa->schema->x_position || j == pa->schema->y_position )

This mi CMake Build configuration(GHT tags/v0.1.0):
-- The C compiler identification is GNU 5.4.0
-- The CXX compiler identification is GNU 5.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting POINTCLOUD build type - RelWithDebInfo
-- Looking for include file stdint.h
-- Looking for include file stdint.h - found
-- Looking for include file getopt.h
-- Looking for include file getopt.h - found
-- Found LibXml2: /usr/lib/x86_64-linux-gnu/libxml2.so (found version "2.9.3") 
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.8") 
-- Found CUnit: /usr/lib/x86_64-linux-gnu/libcunit.so  
-- Found LibGHT: /usr/local/lib/libght.so  
-- Found LazPerf: /usr/local/include  
-- Configuring done
-- Generating done
-- Build files have been written to: /home/carlo/Downloads/UbuntuInstalls/pgpointcloud/pointcloud-build

animecomico:
I see that in ../pointcloud/lib/pc_api.h in PCSCHEMA not exists x_position.
I  see exists pointer to xdim and ydim.

So i change to xdim->position and ydim->position where exists x_position and y_position.

This fix problem with x-position and y_postion.

Another error is with ../pointcloud/lib/cunit/cu_pc_patch.c, error by many arguments in the functions when in ../pointcloud/lib/pc_api.h the function is defined with 3 arguments.

[...]

PR #177, opened 2017-05-11
Build and test with libght in Travis

elemoine:
#175 has reported that we've broken the build when libght is enabled. The problem is that we don't currently build with libght enabled on Travis. This PR is a resurrection of #134 and suggests to build and run unit tests with libght enabled on Travis.

Depends on #176, which should be merged first.

### DIFF
--- a/.install-libght.sh
+++ b/.install-libght.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+set -ex
+git clone https://github.com/pramsey/libght.git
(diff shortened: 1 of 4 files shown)


mbredif:
Great. Could we check that it still builds without lazperf as well ?

elemoine:
> Could we check that it still builds without lazperf as well ?

Yes. I suggest that we do it with a separate PR, if necessary.

elemoine:
I'll merge this in the coming days if I don't get any push back.

PR #183, opened 2017-05-31
doc: fix README&NEWS, pc_boundingdiagonal doesn't exists

ldgeo:


elemoine:
While you're at it, can you plase add a commit that removes the extra characters in the PC_BoundingDiagonalGeometry(pcpatch) example:

```
(11 rows)gq
```

Thanks.

ldgeo:
done

PR #184, opened 2017-06-02
Mention --with-pgconfig in README

elemoine:


elemoine:
I will merge this when Travis is happy.

PR #185, opened 2017-06-12
Make hexbytes_from_bytes perform better

elemoine:
This PR does two things.

1. It greatly improves the performance of the `hexbytes_from_bytes` performance by avoiding calling `snprintf` for each byte. The implementation is now [similar](https://github.com/postgis/postgis/blob/svn-trunk/liblwgeom/lwout_wkb.c#L34-L58) to PostGIS.
1. It adds a `pc_` prefix to the `bytes_from_hexbytes`, `hexbytes_from_bytes` and `wkb_get_pcid` functions to avoid conflicts with other libraries like PostGIS.

Together, these two commits fix a major performance bug. See the [discussion](https://www.postgresql.org/message-id/flat/49f41956-f50b-9551-1943-6e12ed12fa29%40oslandia.com#49f41956-f50b-9551-1943-6e12ed12fa29@oslandia.com) on the pgsql-general mailing list and the [test case](https://github.com/elemoine/pgpc-perf) for more detail.

### DIFF
--- a/lib/cunit/cu_pc_patch.c
+++ b/lib/cunit/cu_pc_patch.c
@@ -139,7 +139,7 @@ test_patch_hex_in()
 	double d;
 	char *str;
 	size_t hexsize = strlen(hexbuf);
-	uint8_t *wkb = bytes_from_hexbytes(hexbuf, hexsize);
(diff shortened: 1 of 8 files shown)


elemoine:
Important enough to be merged sooner rather than later :)

Issue #71, opened 2015-04-14
Memory corruption in schema parsing


strk:
Memory gets corrupted by schema parser. 
As a consequence, backend crashes happens some time later in a seamingly random way...

The following valgrind report is with master branch, as of fc409ca842065d0a66d8948859d124b57f084da2

I hadn't tested 1.0.x

```
2015-04-14 10:28:52.060 CEST  16738   STATEMENT:  select f.schema::xml from pointcloud_formats f, pointcloud_columns c where "table"='toastable' and "column" = 'p' and f.pcid = c.pcid;

==16738== Invalid read of size 8
==16738==    at 0x4E68764: xmlCleanupCharEncodingHandlers (in /usr/lib/x86_64-linux-gnu/libxml2.so.2.9.1)
==16738==    by 0x4E87ED8: xmlCleanupParser (in /usr/lib/x86_64-linux-gnu/libxml2.so.2.9.1)
==16738==    by 0x851EE14: pc_schema_from_xml (pc_schema.c:597)
==16738==    by 0x8516E51: pc_schema_from_pcid_uncached (pc_pgsql.c:272)
==16738==    by 0x8517058: pc_schema_from_pcid (pc_pgsql.c:393)
==16738==    by 0x851657C: pcpatch_get_stat (pc_access.c:825)
==16738==    by 0x5A4081: ExecMakeFunctionResultNoSets (execQual.c:2027)
==16738==    by 0x5A402C: ExecMakeFunctionResultNoSets (execQual.c:2001)
==16738==    by 0x5A402C: ExecMakeFunctionResultNoSets (execQual.c:2001)
==16738==    by 0x5A9FDC: ExecProject (execQual.c:5295)
==16738==    by 0x5AA3C9: ExecScan (execScan.c:207)
==16738==    by 0x5A3007: ExecProcNode (execProcnode.c:400)
==16738==  Address 0x81cbfd8 is 7,352 bytes inside a block of size 8,192 free'd
==16738==    at 0x4C2BDEC: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==16738==    by 0x77683A: AllocSetDelete (aset.c:551)
==16738==    by 0x777B87: MemoryContextDelete (mcxt.c:190)
==16738==    by 0x5ABED2: FreeExprContext (execUtils.c:360)
==16738==    by 0x5ABF20: FreeExecutorState (execUtils.c:192)
[...]

PR #194, opened 2018-02-26
Remove the generic cache

elemoine:
This is an attempt to fix a memory corruption reported with https://github.com/pgpointcloud/pointcloud/issues/71#issuecomment-368228150.

This comes from a patch I wrote a long time ago, but that I didn't create a PR for. I just tested it, and it seems to fix the problem.


elemoine:
I think that's a good change. Does anywone want to look at this? Otherwise I'll go ahead and merge it.

PR #202, opened 2018-04-27
Drop CMake support for building

### LINKED ISSUE #200: Drop cmake
Both the autotools and cmake can currently be used for building Pointcloud. Maintaining both is a bit of a burden.

We use the autotools on Travis, so autotools is what's tested currently. Also, @robe2 also uses the autotools for building packages for Windows.

So the easiest path is to drop cmake. Are there objections? If not I can create a PR for this.

elemoine:
Fixes #200.


elemoine:
@pblottiere do you mind having a quick look at this please?

CMakeLists.txt
```diff
@@ -1,157 +0,0 @@
...
-endif (LAZPERF_FOUND)
-
-#------------------------------------------------------------------------------
-# define targets for "check" and "installcheck"
-
-if (WITH_TESTS)
-    add_custom_target(check COMMAND ${PROJECT_BINARY_DIR}/lib/cunit/cu_tester)
-
-    find_program(PG_REGRESS pg_regress ${PGSQL_PKGLIBDIR}/pgxs/src/test/regress/ REQUIRED)
-    add_custom_target(installcheck COMMAND ${PG_REGRESS} --inputdir=${PROJECT_SOURCE_DIR}/pgsql --outputdir=${PROJECT_SOURCE_DIR}/pgsql pointcloud pointcloud_columns schema)
```
pblottiere: Do we have something similar with autotools?

elemoine: https://github.com/pgpointcloud/pointcloud/blob/master/Makefile#L10-L11

pblottiere: OK, good to know

pblottiere:
Thanks for cleaning @elemoine!

elemoine:
@mbredif are you also ok with this change?

Here are the reasons for this change:

1. We don't want to keep both cmake and autotools (maintenance burden)
2. Our Windows packager (@robe2) relies on autotools and doesn't use cmake at all
3. We use autotools on Travis

mbredif:
Go for it ! 
I was using cmake because I have more experience with it than with autotools but your arguments are convincing, and I admit we already experienced sync issues between the 2 build systems.
So I will just have to learn a bit about autotools :grinning: ...

PR #205, opened 2018-04-27
Update README.md

mbredif:
typo fixing (PC_EnvelopeBinary -> PC_EnvelopeAsBinary)

Can't we say 1.2 instead of 2.0 ?


elemoine:
> Can't we say 1.2 instead of 2.0 ?

Following [semver](https://semver.org/) "MINOR version when you add functionality in a backwards-compatible manner". In this case removing `PC_Envelope` is not a backwards-compatible change, hence 2.0 rather than 1.x here.

PR #206, opened 2018-04-30
Fix pc_patch_compress leak (take 2)

elemoine:
This PR replaces https://github.com/pgpointcloud/pointcloud/pull/112.

I've taken @achidlow's commits, squashed them together, and added three additional commits. My second commit renames the pc_patch_common_free function to pc_patch_free_stats, which is more explicit.

@mbredif can you please take a final look at this before I merge it and release 1.1.0?

### DIFF
--- a/lib/pc_patch_uncompressed.c
+++ b/lib/pc_patch_uncompressed.c
@@ -10,6 +10,7 @@
 *
 ***********************************************************************/
 
+#include <assert.h>
 #include "pc_api_internal.h"
 #include "stringbuffer.h"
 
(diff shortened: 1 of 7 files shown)


elemoine:
Closes #111.

lib/pc_patch_uncompressed.c
```diff
@@ -215,6 +216,8 @@ pc_patch_uncompressed_compute_extent(PCPATCH_UNCOMPRESSED *patch)
 void
 pc_patch_uncompressed_free(PCPATCH_UNCOMPRESSED *patch)
 {
+	pc_patch_free_stats((PCPATCH*) patch);
+
```
mbredif: As you removed the check in `pc_patch_free_stats`, maybe you can add `assert(patch);` here, just to be consistent across patch types

elemoine: Good point. I'll add that.

Nice picture by the way ;)

elemoine: New commit added.

Issue #196, opened 2018-03-02
libght breaks schema parsing


gabri94:
Enabling libght during the compilations of pgpointcloud breaks the schema parsing causing a fatal error in the database.
see #71.

pblottiere:
GHT support has been removed.

Issue #35, opened 2015-02-04
patch fails to compress as GHT


strk:
The patch reported in #31 fails to compress as GHT.
A reduced dataset, of 2 pcpoint records only:

```
 0118000000B0B9FFFF3AEEFFFF00000000FFFF
 0118000000B0B9FFFF6CEEFFFF000000000000
```

Constructing the patch from literal values work fine, but from a table, it fails.

strk:
Self contained test:
```
BEGIN;
COPY pointcloud_formats (pcid, srid, schema) FROM stdin;
[...]

PR #214, opened 2018-06-11
Remove GHT support

pblottiere:
Hi everyone,

This PR follows this [thread](https://lists.osgeo.org/pipermail/pgpointcloud/2018-June/000289.html) saying that `GHT is dead. It's a bad fit for the db environment.`.

So I propose to remove the GHT support from pointcloud. Do you agree?

If this PR is merged, I'll take care about updating the [PDAL plugin](https://github.com/PDAL/PDAL/tree/master/plugins/pgpointcloud/io) accordingly.

Moreover, the next issues should be closed : 
- https://github.com/pgpointcloud/pointcloud/issues/35
- https://github.com/pgpointcloud/pointcloud/issues/196


elemoine:
If we remove GHT support we'll at least need to do the same in PDAL.

pblottiere:
> If we remove GHT support we'll at least need to change PDAL's pgpointcloud writer to reflect that change.

Indeed! That's what I meant by saying : 

> If this PR is merged, I'll take care about updating the PDAL plugin accordingly.

elemoine:
Sorry I hadn't noticed your comment. We're inline.

.travis.yml
```diff
@@ -26,7 +25,7 @@ addons:
 
 script:
   - eclint check * */* */cunit/*
-  - ./tools/build-install.sh --with-lazperf=/usr/local --with-libght=/usr/local && make check
-  - ./tools/build-install.sh --with-lazperf=/usr/local --without-libght && ./tools/valgrind.sh
+  - ./tools/build-install.sh --with-lazperf=/usr/local && make check
+  - ./tools/build-install.sh --with-lazperf=/usr/local && ./tools/valgrind.sh
```
elemoine: Now that we don't build with libght we no longer need to run build-install.sh twice.

pblottiere: Fixed

pblottiere:
Thanks for your review @elemoine 

> I'd like that we tag v1.1.1 before merging this. And we could create v1.2.0 when the GHT support removal is merged.

I fully agree :+1:!

elemoine:
Let's merge this.

PR #215, opened 2018-06-11
Fix bug in pc_patch_uncompressed_from_lazperf

elemoine:
There is a major bug in `pc_patch_uncompressed_from_lazperf` preventing `PC_Uncompress` from working on LAZ-compressed patches.

This is what we currently get when calling `PC_Uncompress` on a LAZ-compressed patch:

```sql
lopocs=# select pc_uncompress(points) from inrap where id=1;
ERROR:  pc_patch_uncompressed_deserialize: calculated patch data sizes don't match (0 != 16000)
```

The bug is that `pc_patch_uncompressed_from_lazperf` doesn't set `datasize` on the uncompressed patch. This PR fixes it.


elemoine:
I suggest to create a v1.1.1 release when this is merged.

pblottiere:
@elemoine Do you think that we could add a unit test covering this issue?

> I suggest to create a v1.1.1 release when this is merged.

:+1:

elemoine:
> @elemoine Do you think that we could add a unit test covering this issue?

Go call! I extended one of the existing lazperf unit tests, and I added a `SELECT PC_Uncompress` query to the lazperf SQL tests.

pblottiere:
> I extended one of the existing lazperf unit tests, and I added a SELECT PC_Uncompress query to the lazperf SQL tests.

It's perfect, thanks @elemoine!

PR #217, opened 2018-06-12
Improve doc for PC_FilterBetween

elemoine:


elemoine:
Non-functional change, merging.

PR #218, opened 2018-06-12
marking most functions as PARALLEL SAFE

tsl-karlp:
Feedback appreciated. :) . I didn't mark the `upgrade` function as PARALLEL SAFE.

Not sure if I got the tests correct, too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -44,7 +44,7 @@ if test "x${GITCMD}" = "xyes" -a "x${DOTGITDIR}" = "xyes"; then
 else
     GIT_COMMIT_HASH=
 fi
-POINTCLOUD_VERSION="$(cat Version.config)$GIT_COMMIT_HASH"
+POINTCLOUD_VERSION="$(cat Version.config)"
 AC_SUBST([POINTCLOUD_VERSION])
 AC_DEFINE_UNQUOTED([POINTCLOUD_VERSION], ["$POINTCLOUD_VERSION"], [Pointcloud version])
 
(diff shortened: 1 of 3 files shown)


elemoine:
Aggregates could  probably be marked PARALLEL SAFE as well.

One issue is that, in their current forms, the changes break the compatibility with PostgreSQL < 9.6. PostGIS pre-processes the [postgis.sql.in](https://github.com/postgis/postgis/blob/d5d26781f765598ea0639e454349a025ceec09b3/postgis/postgis.sql.in) files to address that problem.

elemoine:
> @elemoine : Is compatibility with PostgreSQL < 9.6 a requirement for pointcloud?

I think it would be a shame if we did not support PostgreSQL < 9.5 just because of `PARALLEL SAFE`.

elemoine:
FWIW PostGIS uses the cpp preprocessor for that. See https://github.com/postgis/postgis/blob/68fc1fec36dc48ac507dc4c8ee3e3f102a68243a/postgis/Makefile.in#L208-L210.

tsl-karlp:
Unfortunately, this was the smallest lift for us, given our time.  Feel free to take over. :)

tsl-karlp:
@elemoine : I brought over some code from postgis, which detects PG versions at the `./configure` phase, and autoconf takes care of the rest.  All the tests pass.  

Is there something missing from this PR?

elemoine:
Thanks for the work on this @tsl-karlp. This is great. I'll have a look at your patch next week.

configure.ac
```diff
@@ -325,6 +325,42 @@ fi
...
+dnl Starting with PostgreSQL 10, major is the new minor
+dnl This is to prevent things like 10.31 ranking higher than 11.0
+dnl===================================================================
+PGSQL_MINOR_VERSION=0
+fi
+
+
+AC_MSG_RESULT([checking PostgreSQL version... $PGSQL_FULL_VERSION])
+
+POSTGIS_PGSQL_VERSION="$PGSQL_MAJOR_VERSION$PGSQL_MINOR_VERSION"
```
elemoine: PostGIS [uses](https://github.com/postgis/postgis/blob/68fc1fec36dc48ac507dc4c8ee3e3f102a68243a/configure.ac#L453) the `POSTGIS_` prefix because it's PostGIS :) Let's not use the `POSTGIS_` prefix in the Pointcloud extension.

tsl-karlp:
Superseded by #227 .

PR #220, opened 2018-06-14
Add PC_MakePatch(pcid integer, values float8[])

elemoine:
Add PC_MakePatch to create a patch from a pcid and an array of float8 values.

Supersedes #219.


elemoine:
Thanks @tsl-karlp for the patch.

Issue #243, opened 2019-10-14
pointcloud doesn't compile against PostgreSQL 12


robe2:
PostgreSQL 12 came out very recently and I'm packaging for windows.  When I try to compile pgpointcloud 1.2 (or master) against PostgreSQL 12, I get these errors.  Works fine compiling against lower versions of PostgreSQL.

> 
> make[1]: Entering directory '/projects/postgis/pointcloud/pointcloud/pgsql'
> 
> gcc -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -g -O2  -I../lib -I. -I./ -IE:/jenkins/postgresql/rel/pg12w64gcc81/include/server -IE:/jenkins/postgresql/rel/pg12w64gcc81/include/internal  -I/projects/zlib/rel-zlib-1.2.11w64gcc81/include  -I/projects/libxml/rel-libxml2-2.9.9w64gcc81/include -I./src/include/port/win32 -DEXEC_BACKEND -I/projects/libxml/rel-libxml2-2.9.9w64gcc81/include/libxml2  -IE:/jenkins/postgresql/rel/pg12w64gcc81/include/server/port/win32 -DWIN32_STACK_RLIMIT=4194304  -c -o pc_inout.o pc_inout.c
> 
> In file included from pc_inout.c:12:
> pc_pgsql.h:81:46: error: unknown type name 'FunctionCallInfoData'; did you mean 'FunctionCallInfoBaseData'?
>  PCSCHEMA* pc_schema_from_pcid(uint32_t pcid, FunctionCallInfoData *fcinfo);
>                                               ^~~~~~~~~~~~~~~~~~~~
>                                               FunctionCallInfoBaseData
> pc_pgsql.h:93:66: error: unknown type name 'FunctionCallInfoData'; did you mean 'FunctionCallInfoBaseData'?
>  PCPOINT* pc_point_from_hexwkb(const char *hexwkb, size_t hexlen, FunctionCallInfoData *fcinfo);
>                                                                   ^~~~~~~~~~~~~~~~~~~~
>                                                                   FunctionCallInfoBaseData
[...]

PR #244, opened 2019-10-15
Support for PostgreSQL 12 references #243

robe2:
I copied this from fix we did in PostGIS. 

Note I confirmed it compiles but since I don't have lazperz tests often fail for me.

https://trac.osgeo.org/postgis/ticket/4313


elemoine:
Looks good to me. Thanks.

PR #246, opened 2019-11-14
add support for binary protocol for pcpatch

cyrilleberger:
This is an updated version of #85, with a test and update.

Two comments
* the test requires /tmp/__pgpointcloud_pa_test_file__ to be writable
* the update to pg_type to install the pcpatch_send/pcpatch_recv functions is run even on new installation, which is unneeded (but harmless, as far as I know)

No idea on how to do differently and if it is a big problem.


cyrilleberger:
I was wondering, is anything else needed?

cyrilleberger:
Not entirely sure why the change I made would cause the errors, but  if there is ever any interest in merging I could have a look.

cyrilleberger:
Formatting fixed, and updated to the latest master. "make check" is still passing.

PR #257, opened 2020-01-08
Don't overwrite CXXFLAGS from environment

df7cb:
Spotted by Debian's build log hardening checker (blhc).


pblottiere:
Thanks @df7cb :+1:

Issue #256, opened 2020-01-08
pointcloud regression tests fail on 32 bit


df7cb:
The regression tests fail on Debian/i386:
```
9.4/regress ...
12:05:51 *** pgsql/regression.diffs ***
12:05:51 + echo *** pgsql/regression.diffs ***
12:05:51 + cat pgsql/regression.diffs
12:05:51 *** /tmp/autopkgtest.efWSUZ/tree/pgsql/expected/pointcloud_pg9.out	2020-01-07 20:42:28.000000000 +0000
12:05:51 --- /tmp/autopkgtest.efWSUZ/tree/pgsql/results/pointcloud_pg9.out	2020-01-08 11:05:49.799218922 +0000
12:05:51 ***************
12:05:51 *** 395,401 ****
12:05:51   SELECT Sum(PC_MemSize(pa)) FROM pa_test_dim;
12:05:51    sum 
12:05:51   -----
12:05:51 !  684
12:05:51   (1 row)
12:05:51   
12:05:51   SELECT Sum(PC_PatchMax(pa,'x')) FROM pa_test_dim;
12:05:51 --- 395,401 ----
12:05:51   SELECT Sum(PC_MemSize(pa)) FROM pa_test_dim;
12:05:51    sum 
12:05:51   -----
12:05:51 !  668
12:05:51   (1 row)
12:05:51   
12:05:51   SELECT Sum(PC_PatchMax(pa,'x')) FROM pa_test_dim;
12:05:51 ***************
12:05:51 *** 431,437 ****
12:05:51   SELECT Sum(PC_MemSize(pa)) FROM pa_test_dim;
12:05:51    sum  
12:05:51   ------
12:05:51 !  8733
12:05:51   (1 row)
12:05:51   
12:05:51   SELECT Max(PC_PatchMax(pa,'x')) FROM pa_test_dim;
12:05:51 --- 431,437 ----
12:05:51   SELECT Sum(PC_MemSize(pa)) FROM pa_test_dim;
12:05:51    sum  
12:05:51   ------
12:05:51 !  8713
12:05:51   (1 row)
12:05:51   
12:05:51   SELECT Max(PC_PatchMax(pa,'x')) FROM pa_test_dim;
```

All Debian/Ubuntu releases are affected.

pblottiere:
Thanks for the report @df7cb, I'm going to take a look asap.

pblottiere:
Hi @df7cb,

[...]

PR #260, opened 2020-03-29
Fixes regression tests on 32-bit

pblottiere:
I (finally) took a look on the issue raised [here](https://github.com/pgpointcloud/pointcloud/issues/256) about failing tests on 32-bit architecture.

Actually it's due to the underlying `sizeof(SERIALIZED_PATCH)` call in `PC_MemSize(ppcpatch)` function. Let's consider the next code which is a minimalist way to reproduce the error:

``` c
typedef struct {
  double xmin;
  double xmax;
  double ymin;
  double ymax;
} PCBOUNDS;

typedef struct {
  uint32_t size;
  uint32_t pcid;
  uint32_t compression;
  uint32_t npoints;
  PCBOUNDS bounds;
  uint8_t data[1];
} SERIALIZED_PATCH;

int main() {
  printf("%d\n", sizeof(B));
}
```
The corresponding results are:

- on 32-bit: `52` bytes
- on 64-bit: `56` bytes

So I propose to add a `padding` member to the `SERIALIZED_PATCH` structure in order to preserve a proper 8 bytes alignment:

``` c
typedef struct {
  uint32_t size;
  uint32_t pcid;
  uint32_t compression;
  uint32_t npoints;
  PCBOUNDS bounds;
  uint8_t data[1];
  uint32_t padding; // not used, for 8 bytes alignment only (see #256)
} SERIALIZED_PATCH;
```

This way, we have a size of `56` bytes both on 32-bit and 64-bit architecture.


pblottiere:
Hi @strk and @elemoine,

I'm not very used to fix this kind of issue so can you give me your opinion on this please? And let me know if you have a better idea :).

Thanks :+1:

strk:
I believe postgresql headers have a MEMALIGN or similar macro to deal with these cases.
I think it would be better to use those rather than adding manual padding

strk:
It was MAXALIGN (or INTALIGN or other niceties): https://doxygen.postgresql.org/c_8h.html#a286221fe52f4de1843b8ffdae71a97db

pblottiere:
Hi @strk,

> It was MAXALIGN (or INTALIGN or other niceties):

Nice, thank you for the tip, that's what I need :+1:!

However it doesn't work out of the box. Actually MAXALIGN and others are based on the TYPEALIGN macro which uses some constants to determine the n-bytes alignment. On 32-bit Debian, these constants are:

- ALIGNOF_SHORT: 2
- ALIGNOF_INT: 4
- ALIGNOF_LONG: 4
- ALIGNOF_DOUBLE: 4
- MAXIMUM_ALIGNOF: 4
- ALIGNOF_BUFFER: 32
- ALIGNOF_LONG_LONG_INT: 4

So even if I use `MAXALIGN(sizeof(SERIALIZED_PATCH))`, I still have a 4 bytes alignment. 

But I can directly use `TYPEALIGN(8, sizeof(SERIALIZED_PATCH))`. This way everything seems fine on 32-bit and 64-bit architecture.

@str what do you think? It's still much better than adding an empty padding member in the structure...

strk:
Not sure why you need 8bytes if those macros say you don't.
So maybe you nedd ALIGNOF_BUFFER ?
I'd ask PostgreSQL hackers...

pblottiere:
@strk one last thing. Travis status is not linked to PR since 2 weeks now.

It happened on some personal projects too and I had to disable a new option in Travis to make it work again (I'm not sure it's the same issue but it seems so):

Untitled

I'd do it but I don't have enough rights/privileges on the pgpointcloud organization. Can you take a look please? Because in the meantime, I don't/can't merge PR.

Thanks again.

strk:
@pgpointcloud/owners how about adding @pblottiere in the team having admin privileges ?

elemoine:
> @pgpointcloud/owners how about adding @pblottiere in the team having admin privileges ?

Definitely.

pblottiere:
> how about adding @pblottiere in the team having admin privileges ?

Thanks @strk :).

PR #261, opened 2020-03-29
Fix Travis

pblottiere:


strk:
How about dump Travis for GitHub-Actions ? They are both proprietary but one is more integrated...

pblottiere:
> How about dump Travis for GitHub-Actions ? 

Indeed, it could be the right time. I take a look in the coming days.

pblottiere:
See https://github.com/pgpointcloud/pointcloud/pull/262

PR #290, opened 2022-02-07
Start github actions one time only

pblottiere:


pblottiere:
@lbartoletti (GitHub actions are still obscure for me)

.github/workflows/bionic_postgres12_postgis25.yml
```diff
@@ -1,6 +1,10 @@
 name: "[ubuntu-18.04] PostgreSQL 12 and PostGIS 2.5"
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
lbartoletti, line 7: If I understand, you don't allow running the CI for PR?

pblottiere: Before this PR each action was started twice for each push (because it's a PR and because there's a push on the branch). Now actions are started only once when you open the PR, for each commit or when the PR is merged on master (like QGIS CI).

PR #295, opened 2022-02-08
Move install-lazperf script to tools dir

pblottiere:


pblottiere:
@lbartoletti

.github/scripts/pgpointcloud.sh
```diff
@@ -2,9 +2,9 @@
 
 set -e
 
-./tools/build-install.sh
-sh .install-lazperf.sh
-./tools/build-install.sh --with-lazperf=/usr/local
+sh ./tools/build_install.sh
+sh ./tools/install_lazperf.sh
+sh ./tools/build_install.sh --with-lazperf=/usr/local
```
lbartoletti, line 7: This makes me think that we should embed lazperf in a 3rdparty-like dir. Since laz-perf on system can be various and pgpointcloud requires laz-perf 1

lbartoletti (approved):

.github/scripts/pgpointcloud.sh
pblottiere: I'd probably be in favor of supporting new version of lazperf instead.

Issue #197, opened 2018-04-18
pg_dump fails


dtaelman:
Hi all,

Performing a backup of my postgreSQL databases (using PostgreSQL 10.3) with the pointcloud-extension enabled fails and generates the following errors:
- pg_dump: Dumping the contents of table "points" failed: PQgetResult() failed.
- pg_dump: Error message from server: ERROR:  relation "pointcloud_formats" does not exist
- LINE 1: select schema, srid from pointcloud_formats where pcid = 1
- QUERY:  select schema, srid from pointcloud_formats where pcid = 1
- pg_dump: The command was: COPY public.points (id, pt) TO stdout

Does anyone have an idea what causes the pg_dump to fail and how to solve the issue?

elemoine:
Can you please provide the full pg_dump command line that you're trying to use? Thanks.

dtaelman:
The full pg_dump command is:
pg_dump --file "/Users/devitaelman/Desktop/test.sql" --host "localhost" --port "5432" --username "devitaelman" --verbose --format=p “lidartest”

Thanks
Devi

trey-stafford:
I have also experienced this issue with postgresql 9.6.8, pointcloud v1.1.0:

```
# pg_dump -U $DB_NAME -U $DB_USER -F p --file /share/appdata/dump/db.dump

g_dump: Dumping the contents of table "atm1b" failed: PQgetResult() failed.
pg_dump: Error message from server: ERROR:  relation "pointcloud_formats" does not exist
LINE 1: select schema, srid from pointcloud_formats where pcid = 1
                                                     ^
QUERY:  select schema, srid from pointcloud_formats where pcid = 1
pg_dump: The command was: COPY public.atm1b (id, pa) TO stdout
```

trey-stafford:
Note that I _can_ dump the pointcloud_formats table:
```
pg_dump -U $DB_USER -F p --file /share/appdata/dump/db.dump -t pointcloud_formats
```

gabri94:
I have the very same error trying to upgrade from postgres 9.5 to postgres 9.6
see #211

trey-stafford:
I have come back to this issue again recently, and have tested postgresql versions v9.6.7 through v9.6.10.

The problem with pg_dump appears in postgresql v9.6.8 through v9.6.10.

pg_dump does work with postgresql v9.6.7.

foggel:
[...]

Issue #211, opened 2018-05-25
Errore while upgrading from postgres9.5 to postgres 9.6


gabri94:
Hi, i'm trying to upgrade my cluster from version 9.5 to 9.6.
I'm using postgis 2.2 in the old one and 2.3 in the new one and pointcloud 1.1 is installed in both.
When i run `sudo pg_upgradecluster 9.5 main -m upgrade` to upgrade the cluster
I get this error:
```
_restore: [archiver (db)] could not execute query: ERROR:  function pc_envelopeasbinary(public.pcpatch) does not exist
LINE 2:   SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))
                                 ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:  
                SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))

CONTEXT:  SQL function "pc_envelopegeometry" during inlining
    Command was: 
-- For binary upgrade, must preserve pg_class oids
SELECT pg_catalog.binary_upgrade_set_next_index_pg_class_oid('8872117'::...
```
 
instead, by running  `sudo pg_upgradecluster 9.5 main` i get this other one:
```
Disabling connections to the new cluster during upgrade...
Roles, databases, schemas, ACLs...
 set_config 
------------
 
(1 row)

ERROR:  function pc_envelopeasbinary(public.pcpatch) does not exist
LINE 2:   SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))
                                 ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:  
                SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))

CONTEXT:  SQL function "pc_envelopegeometry" during inlining
ERROR:  function pc_envelopeasbinary(public.pcpatch) does not exist
LINE 2:   SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))
                                 ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:  
                SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))

CONTEXT:  SQL function "pc_envelopegeometry" during inlining
ERROR:  function pc_envelopeasbinary(public.pcpatch) does not exist
LINE 2:   SELECT ST_GeomFromEWKB(PC_EnvelopeAsBinary($1))
                                 ^
[...]

PR #301, opened 2022-02-10
Fix dump and restore

pblottiere:
Fixes https://github.com/pgpointcloud/pointcloud/issues/197 https://github.com/pgpointcloud/pointcloud/issues/211

TODO:
- [x] add tests

### DIFF
--- a/.github/scripts/test_dump_restore.sh
+++ b/.github/scripts/test_dump_restore.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#! /usr/bin/env bash
 
 set -e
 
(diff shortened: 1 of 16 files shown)


lbartoletti:
Tested locally with sql and custom dump on a tiny dataset and it's OK. Thanks!

.github/scripts/test_dump_restore.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
```
lbartoletti, line 1:
```suggestion
#! /usr/bin/env bash
```

Avoid gnuism

lbartoletti (approved):
Restore tested and it seems OK!
Thanks

PR #302, opened 2022-02-10
Add CI for PG 10/11 with PostGIS 2.5

pblottiere:
Note: specific tests for PG 9.6 are back again on Ubuntu 18.04 (instead of Ubuntu 16.04).


pblottiere:
@lbartoletti

lbartoletti:
Why PG 9.6? It's not supported by PostgreSQL since November 11, 2021 https://www.postgresql.org/support/versioning/

Otherwise, it's ok.

pblottiere:
> Why PG 9.6?

Because I have some specific test cases for PG 9.6. But yes, I can just drop everything about the PG 9.6 release (CI and tests).

PR #312, opened 2022-05-25
Don't call SPI_finish() when erroring out

df7cb:
SPI_connect() opens a new memory context, which is then destroyed by
SPI_finish(). Since we are invoking quote_qualified_identifier()
inbetween, which pallocs its result in that memory context, we cannot
use "formats" in the error messages because that pointer has already
been freed by then. Fix by dropping all SPI_finish() calls in paths that
elog(ERROR) which do handle the memory cleanup anyway.


pblottiere:
Thanks a lot @df7cb :+1:

Issue #314, opened 2022-06-04
pgpointcloud doesn't compile against PostgreSQL 15beta1


robe2:
I ran into two issues compiling pgpointcloud under mingw64 with PG15.
1) First is macro expansion doesn't handle beta.
2) pg_atoi was removed in PG15.  

Take a look at  https://github.com/MobilityDB/MobilityDB/issues/218

I'll take a stab at patching these up later if not addressed yet.

```
pc_pgsql.h:97:5: note: in expansion of macro 'PGSQL_VERSION'
 #if PGSQL_VERSION < 120
     ^~~~~~~~~~~~~
../lib/pc_config.h:4:23: error: invalid suffix "PostgreSQL" on integer constant
 #define PGSQL_VERSION 15PostgreSQL 15beta1
                       ^~~~~~~~~~~~
pc_pgsql.h:118:5: note: in expansion of macro 'PGSQL_VERSION'
 #if PGSQL_VERSION < 120
     ^~~~~~~~~~~~~
../lib/pc_config.h:4:36: error: missing binary operator before token "15beta1"
 #define PGSQL_VERSION 15PostgreSQL 15beta1
                                    ^~~~~~~
pc_pgsql.h:118:5: note: in expansion of macro 'PGSQL_VERSION'
 #if PGSQL_VERSION < 120
     ^~~~~~~~~~~~~
pc_inout.c: In function 'pc_typmod_in':
pc_inout.c:399:13: warning: implicit declaration of function 'pg_atoi'; did you mean 'pg_ltoa'? [-Wimplicit-function-declaration]
    typmod = pg_atoi(s, sizeof(int32), '\0');
             ^~~~~~~
             pg_ltoa
make[1]: *** [<builtin>: pc_inout.o] Error 1
make[1]: Leaving directory '/projects/postgis/pointcloud/pointcloud/pgsql'
make: *** [Makefile:4: all] Error 2
```

robe2:
I realized I was compiling against tagged.  So there is only one issue the pg_atoi one.  I'm testing out patch now for that and will do a pull request shortly.

PR #315, opened 2022-06-04
Fix for PG15, closes #314

robe2:
This defines pg_atoi function for PostgreSQL >= 15

PR #318, opened 2022-08-19
PG15 support (with CI)

pblottiere:
Fixes https://github.com/pgpointcloud/pointcloud/issues/314

### DIFF
--- a/.github/scripts/postgresql_postgis.sh
+++ b/.github/scripts/postgresql_postgis.sh
@@ -3,7 +3,7 @@
 set -e
 
 wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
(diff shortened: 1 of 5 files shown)


pblottiere:
Hello @robe2,

I noticed your issue https://github.com/pgpointcloud/pointcloud/issues/314 and the corresponding PR https://github.com/pgpointcloud/pointcloud/pull/315.

Actually, `pg_atoi` is used at a single location so I propose to follow the same approach as in Postgis: https://trac.osgeo.org/postgis/changeset/f1e799c7ee4a76f0d1029f91983f9dd2e05db8fe/git.

Moreover, this PR add PostgreSQL 15 in CI.

Let me know what you think.

robe2:
> 

That works for me.

Issue #282, opened 2021-02-09
Some inputs makes postgres crash


autra:
# Description

Some input makes postgresql crash

# Step to reproduce

- Have pointcloud_postgis installed on postgresql 12 or 13.
- execute `select '01010000000000000000000000'::pcpatch;`
- observe the crash

# Full test

Please find attached a sql script to reproduce that: 

<details>
<summary>sql script</summary>
<pre>

SET client_encoding = 'UTF8';
CREATE EXTENSION IF NOT EXISTS pointcloud WITH SCHEMA public;
CREATE EXTENSION IF NOT EXISTS postgis WITH SCHEMA public;
CREATE EXTENSION IF NOT EXISTS pointcloud_postgis WITH SCHEMA public;

truncate public.pointcloud_formats;

COPY public.pointcloud_formats (pcid, srid, schema) FROM stdin;
1	4326	<?xml version="1.0" encoding="UTF-8"?>\n<pc:PointCloudSchema xmlns:pc="http://pointcloud.org/schemas/PC/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">\n <pc:dimension>\n  <pc:position>1</pc:position>\n  <pc:size>8</pc:size>\n  <pc:description>X coordinate</pc:description>\n  <pc:name>X</pc:name>\n  <pc:interpretation>double</pc:interpretation>\n  <pc:active>true</pc:active>\n </pc:dimension>\n <pc:dimension>\n  <pc:position>2</pc:position>\n  <pc:size>8</pc:size>\n  <pc:description>Y coordinate</pc:description>\n  <pc:name>Y</pc:name>\n  <pc:interpretation>double</pc:interpretation>\n  <pc:active>true</pc:active>\n </pc:dimension>\n <pc:dimension>\n  <pc:position>3</pc:position>\n  <pc:size>8</pc:size>\n  <pc:description>Z coordinate</pc:description>\n  <pc:name>Z</pc:name>\n  <pc:interpretation>double</pc:interpretation>\n  <pc:active>true</pc:active>\n </pc:dimension>\n <pc:metadata>\n  <Metadata name="compression" type="string">dimensional</Metadata>\n </pc:metadata>\n <pc:orientation>point</pc:orientation>\n <pc:version>1.3</pc:version>\n</pc:PointCloudSchema>\n
\.

drop table if exists test;
CREATE TABLE test(
    id integer NOT NULL,
    pa public.pcpatch(1)
);

[...]

PR #320, opened 2022-08-25
No crash when invalid dimensional patch

pblottiere:
Fixes https://github.com/pgpointcloud/pointcloud/issues/282


pblottiere:
@autra fyi

PR #350, opened 2023-11-18
Let 'make' use CXX configuration variable

nilason:
Let 'make' use CXX configuration variable and the "-std=c++11" flag.


strk:
I think c++11 is fine in 2025, thanks

PR #351, opened 2023-12-22
pgsql: respect compiler choice

barracuda156:
Current makefile ignores compiler choice. Passing CC and CXX values to environment has no effect. In result the build may be broken: https://trac.macports.org/ticket/68867
Fix it, so that the chosen compiler is actually used for all components coherently.


strk:
This should have been fixed by merging #350

PR #354, opened 2024-07-12
Update patchs.rst

Garbage123King:
The formats of max and avg are different from min, which is confusing. Change them to the same format as min.


pblottiere:
Yes indeed!

A very belated thank you :+1:

PR #383, opened 2026-07-11
Fix build against PostgreSQL 19

devrimgunduz:
1. pc_access.c PG19 removed the `bits8` typedef (utils/array.h now uses `uint8` directly, e.g. ARR_NULLBITMAP() returns uint8*). Replace all uses of `bits8` with `uint8` (array_get_isnull(), pcpatch_from_point_array(), pcpatch_from_patch_array(), array_to_cstring_array()).

 2. pc_pgsql.c PG19 added an `int *fgc_flags` output parameter to FuncnameGetCandidates() (8 args instead of 7), used to report why a lookup failed. Add a new `#elif PGSQL_VERSION < 190` branch that keeps the existing PG14-PG18 (7-arg) call, and a PG19+ branch that passes a local `int fgc_flags` to satisfy the new signature. The flags aren't otherwise consumed since this call only wants the OID of a uniquely-named function.

Per https://github.com/pgdg-packaging/pgdg-rpms/issues/213

### DIFF
--- a/pgsql/pc_access.c
+++ b/pgsql/pc_access.c
@@ -126,7 +126,7 @@ Datum pcpoint_get_values(PG_FUNCTION_ARGS)
   PG_RETURN_ARRAYTYPE_P(result);
 }
 
-static inline bool array_get_isnull(const bits8 *nullbitmap, int offset)
+static inline bool array_get_isnull(const uint8 *nullbitmap, int offset)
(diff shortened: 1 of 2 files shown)


lbartoletti (approved):
LGTM, is it ok for you @pblottiere ?

